PR #1362, opened 2025-08-05
Fix asset panel folders resizing handle when scrolled down

### LINKED ISSUE #1277: Assets Folders resizing handle issue
Resizing handle for assets folder, is within scrollable element, so when you scroll down within it, it is not possible to resize that panel anymore, as resizing handle is overflown.

## Solution

Ensure that handle is not within scrollable element.

Image

Maksims:
Fixes #1277

It seems that there is an inherit issue with PCUI container element, if it is set to be scrollable as well as resizable, then scrollable style is set on the parent element, leading to resizing handle to be overflow-cut.
In Editor other panels, like hierarchy panel have multiple containers - one for resizing, and then inside of it one for scrolling.
This PR does the same - it adds a new container in for resizing, and then container with files for scrolling as before.

Also tested that auto-scrolling when dragging - still works.

This also fixes the horizontal scroll issue, below are two screens before and after the change (green highlight for the screen):
image image

### DIFF
--- a/src/editor/assets/asset-panel.ts
+++ b/src/editor/assets/asset-panel.ts
@@ -26,6 +26,7 @@ import { tooltip, tooltipSimpleItem } from '../../common/tooltips.ts';
 import { bytesToHuman, naturalCompare } from '../../common/utils.ts';
 
 const CLASS_ROOT = 'pcui-asset-panel';
+const CLASS_LEFT = `${CLASS_ROOT}-left`;
 const CLASS_FOLDERS = `${CLASS_ROOT}-folders`;
 const CLASS_CURRENT_FOLDER = `${CLASS_ROOT}-current-folder`;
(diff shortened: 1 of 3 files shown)


src/editor/assets/asset-panel.ts
```diff
@@ -417,15 +418,21 @@ class AssetPanel extends Panel {
         this._createTooltip('Open Store', btnStore);
 
         // folders tree view
-        this._containerFolders = new Container({
-            class: CLASS_FOLDERS,
+        this._containerLeft = new Container({
```
willeastcott, line 421: Can you please change the comment above to something like "container for folders treeview container that preserves resize handle when scrolling the folder container"?

Maksims:
> Great stuff! Your first editor PR in 8 years!

I needed a bit of rest 🙏

willeastcott:
> I needed a bit of rest 🙏

LOL. Glad you're back. ❤️

PR #1363, opened 2025-08-05
Asset Panel Type fixes

marklundin:
This PR fixes many of the type error in the `asset-panel.ts` and `element-table-row.ts`

- Converts static members like `AssetPanel.MEMBER` to `static MEMER`
- Declares field declarations for class members with JSdoc types
- Few minor type fixes.

### DIFF
--- a/src/editor/assets/asset-panel.ts
+++ b/src/editor/assets/asset-panel.ts
@@ -315,12 +315,24 @@ class AssetPanel extends Panel {
      */
     _btnDelete;
 
+    /**
+     * @type {Button}
+     */
     _btnBack;
 
+    /**
+     * @type {Button}
+     */
     _btnLargeGrid;
 
+    /**
+     * @type {Button}
+     */
     _btnSmallGrid;
 
+    /**
+     * @type {Button}
+     */
     _btnDetailsView;
 
     /**
--- a/src/editor/assets/asset-panel.ts
+++ b/src/editor/assets/asset-panel.ts
@@ -386,7 +386,7 @@ class AssetPanel extends Panel {
      */
     _hoveredAsset;
 
-    /**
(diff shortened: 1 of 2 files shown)


src/editor/assets/asset-panel.ts
```diff
@@ -262,6 +287,260 @@ class AssetGridViewItem extends GridViewItem {
...
+    static VIEW_SMALL_GRID = 'sgrid';
+
+    static VIEW_DETAILS = 'details';
+
+    static LEGACY_SCRIPTS_ID = LEGACY_SCRIPTS_ID;
+
+    /**
+     * @type {Container}
+     */
+    _containerControls;
```
willeastcott, line 302: Why is this not just:

```ts
    _containerControls: Container;
```

marklundin: Are we happy to start adding TS types? If so, then absolutely! 👍

kpal81xd: Yep we should start converting it all to TS

Maksims: That would be a massive waste of time and effort instead of producing features and fixing actual bugs.

Moving code around without - looks like work, but what is it for the end users?

marklundin: I was hoping the JSDoc types improved the Intellisense issues you found. It's just a search replace to change the jsdoc tags to TS types

willeastcott: Yeah, we don't have any plans to work on that ourselves at the moment (unless someone's working on a particular file for some other reason and makes some TS updates while they're there). If any community members feel like doing TS-related updates, be my guest.

PR #1367, opened 2025-08-05
Updated gizmos for translate/rotate/scale

kpal81xd:
Requires https://github.com/playcanvas/engine/pull/7887

- Adds engine as devDependency for including types
- Replaces translate/rotate/scale gizmos with ones from engine


Maksims:
This is a significant one.
Happy to test it before release, to ensure we don't get degradations in UX.

willeastcott (approved):
I literally cannot believe how simple this PR is. Engine Gizmo system FTW! 🚀

Maksims:
Please do not merge.

After a lot of testing, and running this through couple of other users (artists and developers), we have a few concerns:

### Functional:

1. **Gizmo arrows directions** are changing - this is an issue, as direction intentionally did not change before, to indicate the direction of the main vectors: right, up, forward of an entity. With flipping them around, now it is confusing to know which way the object is facing, is it forward in front or to the back? Is it upside down or is it Y arrow of a gizmo flipped? "Sticky" arrow direction - was intentional, and serves functional purpose. The double-plane is not as large or obtrusive, allowing ease of access to arrows without flips.
2. Translation middle gizmo for moving relative to the camera - is a great addition. But it should work against a plane, and not based on distance to the camera, as moving around on the plane leads it to get closer and further based on proximity to the edges of a screen.
3. **Rotation gizmo has a significant reduction in understanding its functionality** by implementing **unintuitive switch between two modes**. After testing, I realized that when looking more perpendicular to the plane of rotation - it follows current behavior by rotating at the pointer on the plane of rotation. While engaging with gizmo when it is more of at the angle, it switches to other mode, where rotation is simply screen-space dependent. This is hidden behavior and double logic on one element. Extremely bad UX, as people will expect one or the other in different situations, and until you drag around to realize at which mode it is, it will lead to confusion. Don't implement "multi-behaviors" like that please. If you want one or the other, implement a user setting, that switches to one, or the other mode, and let users to make preference. Please favor the existing behavior as a default one, as forcing the change on users - is intrusive.
4. Rotation large circle - is a good addition.
5. Rotation gizmo when looking parallel to the plane - is actually the only case where it is useful to engage with in some cases, with screen-space relative pointer rotation.
6. **Scale gizmo double-axis** - is not as useful as it might seem. If it would be of a consistent scaling - that would be more useful, but having double-axis and scaling them independently, makes it somewhat finnicky to get desirable scale. Consider making them relatively locked.
7. **Guides** - this is a massive reduction in their functionality, and more clutter visually. The purpose of guides - is to allow users to align things relatively. For example when moving object on the XZ plane, you need to know what it is above/below for alignment! New Gizmos only show the guides on dragged axes - which are actually most useless guides when moving an object, as it is stationary, and only provides "distance" and not "alignment" feature. Please bring back the all 3 axes.
8. **Guides occlusion** - new gizmos fully lost that feature. Now the guides don't have occluded lines which are behind something. They are useful in many busy places, and still useful for alignments.
9. New **gizmos are too obstructive when interacting with them** - currently we hide gizmos and show faint lines to allow clarity of view for the user when interacting with gizmos. New gizmos don't even hide themself. Which can be visually on the way of what we trying to see - actual object we are manipulating.
10. **Rotation gizmos when engaged are hiding other axes** - making it harder to understand where other planes are rotated to. The common behavior - rotate entity by Yaw, and by seeing Pitch plane you can easier understand where object is facing now. New gizmos fully hide other planes making it harder to see that.
11. In orthographic view with axis-aligned camera, the rotation surrounding circle - is redundant.
12. Scale gizmo flipping axes - is the same issue as with translation, with additional problem - it is not clear where the positive direction is, so when engaging with flipped scale gizmo axis, it will scale down instead of up when moving into gizmos pointing direction - making it more confusing. Don't flip axes please.

### Visual:

1. **RGB color** - is dull and not apparent. When user has gizmos - this is most interacted element on the viewport. In high-contrast scenes they simply get lost in a noise. Please use common RGB, and not transparent colors.
2. **Hovering color consistency** - it is not clear on what you hover in high-contrast scenes, as gizmo colors are in dull color, and then have insignificant highlight. Please use a single high-contrast color for hovering, to communicate clearly on what is hovered.
3. **Rotation big circle** uses different color for hover - this is unnecessary and inconsistent.
4. **Guides** - their color is somewhat irrelevant.

### Bugs/issues:

1. Multiple times the gizmo became sticky while mouse button was released. In some cases this is true for current gizmos also, worth fixing.
2. Hover mouse entity name popup is not disabled when gizmos are engaged.

### Overall:

1. We believe new gizmos is a **significant downgrade in functionality and visual clarity**.
3. It is unclear why this is changed, what benefit and **what it tries to solve?**
4. If the reason for changing the Editor gizmos is because other places (supersplat) now have their own - then please don't do this. It has no positive benefit for Editor users. If you want consistency, please go the other way - make sure other gizmo is consistent with battle-tested Editor's gizmos functionally and visually.
5. I have not seen or heard a complains about Editor's gizmos - they've been reliable, easy and clear to use gizmos through years. The massive amount of work spent on changing them - is simply not needed and do have reduction in experience.

### Comparison (before/after):

Clarity of translate gizmo:
image image

Translate gizmo dragging by Z axis. Notice missing X guide, and gizmos are not hiding.
image image

Translate gizmo when dragging on two axes (mouse is actually on the right of the gizmo):
image image

Rotation Gizmo (notice that Yaw gizmo is not fully rendered):
image image

Rotation gizmo engaged Yaw axis (notice that it is not clear where Z axis is pointing with new gizmos):
image image

Scale gizmo:
image image

kpal81xd:
@Maksims thanks for your in-depth here are my thoughts on your comments:

### Colors
These were designed with customization in mind (including the hover colors) so can be adjusted accordingly to match (if so required). This idea was to allow the user to customize their gizmos in the editor however they would like

### Axis/plane flipping
This was a recent addition for better accessibility when using the gizmos from obscure angles. Again like the colors this was planned to be added as a user customization feature in the future 

### Translating in facing plane
This is a regression and will be fixed

### Consistent scaling
This is a feature I had missing I will add this (it is present in superspl.at)

### Guides
Actually yes I agree with this - the hover should show the plane but actually movement its better to have all 3. As for the occlusion I had missed this actually and will add 

### Hiding
This was disabled but I will add a setting to reduce opacity when moving for a clearer view

### Bugs
For the sticky mouse issue is this when the mouse leaves the viewport? Which scenarios does this occur in? as for the entity name on hover I will fix this

Maksims:
To reduce the issue of imposed changes on users. It will be the best to match the behavior and look as much as possible by default. As it will reduce the mechanical-memory confusions by users in case of different behavior / perception. And then add settings to customize.

> These were designed with customization in mind (including the hover colors) so can be adjusted accordingly to match (if so required). This idea was to allow the user to customize their gizmos in the editor however they would like

Is it possible to match existing colors by default, and then provide customization by users that need it?

> This was a recent addition for better accessibility when using the gizmos from obscure angles. Again like the colors this was planned to be added as a user customization feature in the future

Can this be disabled by default (not flipping axes by default), and provide a setting in the future for users if they want it?

> This was disabled but I will add a setting to reduce opacity when moving for a clearer view

Can this be enabled by default (gizmos hiding), and then provide a setting later to configure if someone needs it?

> For the sticky mouse issue is this when the mouse leaves the viewport? Which scenarios does this occur in? as for the entity name on hover I will fix this

This actually same for existing gizmos - if you alt+tab to other window and release mouse over other window. Or on Windows do a screenshot with a snippet which has overlay over your screen - it will lead to sticky one.
If it is possible to detect loss of focus on window/tab in these cases - this can help to release gizmos.
Also it is good to check when pointer down while dragging, you can check if there was any other pointers, if no - then it means this is a "sticky situation" :D and can release gizmo before processing new pointerdown.

SaadHaider7:
I checked it and think that the old visuals and behavior were better and the new one should match it and add the extra options as configurable in the settings like Max said.
Users should not be made to re learn the things they are used to using especially the gizmo as its used a lot and they might get frustrated seeing the new one.

willeastcott:
@SaadHaider7 OK, but do you have anything specific you would like to add to @kpal81xd's list above?

SaadHaider7:
The Rotation gizmo, if that can be improved and is less flickery, 

https://github.com/user-attachments/assets/509bd0c5-7f9a-426f-be19-bfd1a99b9d00

Maksims:
I've tested with latest changes, many things **improved** and are behaving almost the same as the current gizmos.

**Rotation gizmos to have two modes of rotation - is a massive downgrade in UX and huge shift in its behavior, which will lead to frustrations for users as it changes current behavior. And it does the changes unintuitively, and often when not expected.
Current behavior is very common across many modeling tools, e.g. Blender.**

### There are couple of issues:

1. Gizmos should be disabled on camera if that camera is used for the viewport. How to replicate: select camera in hierarchy, then click on top-left view in viewport to switch view to that camera. Issue: you will have gizmos rendering within the camera. Expected: gizmos to be disabled.
image

2. Cases of invalid picking, actually is an issue with the current gizmo too, but behavior differs. Basically, if you drag by the plane, and while dragging will move mouse out of the plane, so it goes into infinity, current gizmo picks plane from behind, while new gizmo just "snaps" to some strange floating number. Would be great in such case to snap back to initial value.
3. For translate gizmo, planes are interactive well, but when we have some plane hidden due to steep angle, then axis becomes interactive also, where only a plane interaction is expected:
image  

**4. Rotation gizmo still has two modes of rotation, based on angle to the plane. Please use the current behavior as a default one (rotation angle towards cursor), and provide a settings to customize that. Blender uses same rotation behavior as the current logic, and many other tools. Which is a common behavior for the rotation gizmos. Hidden switches - are an awful UX, and should be avoided at all costs.**
5. While gizmos are active (hovered or interacted with), it should not bring entity name tooltip. Especially during interactions with gizmo.
image

6. In orthographic views, rotation gizmo planes are not fully visible as mentioned by @SaadHaider7 before:
image

7. When you are hovering on rotation gizmos, it renders a guide line, which not sure is useful here.
8. When you are interacting with rotation gizmo, it is useful to see other axes gizmos, especially when using local space as it shows helps to understand the planes of rotations.
9. When interacting with translation gizmo, it would be great to have only active axes as primary highlighted, and others secondary, to indicate which axes we are interacting with. One for single axis, and two for plane interactions.
10. Scale gizmo has some unintended interactive parts, check out video:

https://github.com/user-attachments/assets/5e11ded0-b6c6-4b7c-a852-0c48ee0e11ff

11. When entity has a negative scale, the scale gizmo is missbehaving, making some axes not hoverable, and flips them (is flipping intended?).

https://github.com/user-attachments/assets/e1cd47ed-6eae-457d-b7f7-a96ac0d3ed19

12. If a parent has a negative scale, this confuses world space translations. But that is the same issue with current gizmos. Negative scale is generally always to be avoided anyway, so not a big deal.

13. Colors are much better now, but they still are not simply bright, expected: "255,0,0" for red, "0,255,0" for green and "0,0,255" for blue. This might be some linear workflow challenges here.

kpal81xd:
@Maksims thanks for the testing. Some questions on your findings:

2 - I cannot seem to replicate this - dragging by the plane in translate mode the object still follows depsite moving the mouse outside the windows bounds. What OS and browser are you testing this on?

3 - I believe that is because you are inside the arrow interaction bounds when the plane hides

4 - This was a flag that I had forgotten to enable so should be fixed now

8 - I assume you mean the rotation arcs? If so I agree and will add in the Engine PR as an option

9 - I already anticipated this so I have also have a mode for this 😄 - I will leave it off for now tho just to ensure I get the behavior as close to the existing one first

10, 11, 12 - Broken scaling is known and will be fixed in follow up engine PR

Maksims:
> 2. I cannot seem to replicate this - dragging by the plane in translate mode the object still follows depsite moving the mouse outside the windows bounds. What OS and browser are you testing this on?

It is about dragging stuff within the viewport, but if you visualize the planes that we are dragging around, then there are cases, where on high angles, it is possible to have that plane not to cover whole screen. In such cases we end up dragging gizmos to "outside" of the plane. It is not clear what needs to be done in such case, but I would assume we would want to detect that and not apply transformations in such case, or reset transformation to initial state. If not updating when detecting outside feels ok, and not jumpy - then that option is better. I will record a vid tomorrow to show it better.

kpal81xd:
> > 2. I cannot seem to replicate this - dragging by the plane in translate mode the object still follows depsite moving the mouse outside the windows bounds. What OS and browser are you testing this on?
> 
> It is about dragging stuff within the viewport, but if you visualize the planes that we are dragging around, then there are cases, where on high angles, it is possible to have that plane not to cover whole screen. In such cases we end up dragging gizmos to "outside" of the plane. It is not clear what needs to be done in such case, but I would assume we would want to detect that and not apply transformations in such case, or reset transformation to initial state. If not updating when detecting outside feels ok, and not jumpy - then that option is better. I will record a vid tomorrow to show it better.

I figured out what you meant by moving the gizmo to infinity from a sharp angle. I just keep the gizmo position fixed once the intersection test (close to infinity) and seems to work nicely

Maksims:
> The Rotation gizmo, if that can be improved and is less flickery,
> 
>  bandicam.2025-08-07.20-26-12-910.mp4

This is still an issue with the latest updates.
Orthographical camera - does not show fully the perpendicular axes rotation circle gizmo.

I believe in this view, yellow circle should be disabled, and green circle should be visible fully.
Parallel axes should either be not interactable, or in this very specific case relative mouse mode used (but only here, when it is orthographic camera with exact axis aligned view).

image

Maksims:
The other issue, is with use of more complex hierarchy, non uniform scales and parent rotations.
Current gizmo correctly shows the orientation, and gizmos are draggable as expected, while gizmos in this PR, behave funky and when dragged are doing math wrong.
Video below shows current and new gizmos:

https://github.com/user-attachments/assets/4fbd9b56-3027-4fb8-a0e9-c5f1febd2826

With rotation gizmos, their axes are especially missaligned:

https://github.com/user-attachments/assets/f76c9ee8-86d6-4e46-aa2d-6a0b60a653c3

Maksims:
# Critical!!

If parent has rotation, then selecting any descendant, local gizmo will be not properly transformed.
And interacting with such gizmo does math wrong. Translate, rotate and scale.

https://github.com/user-attachments/assets/a75cd512-dc91-4a1d-87e6-3d997fdce8e2

Maksims:
Is it possible to make a new gizmo look somewhat closer to existing ones (esp. the rectangle bit on translate)? So they don't look that "flat"?

kpal81xd:
> # Critical!!
> If parent has rotation, then selecting any descendant, local gizmo will be not properly transformed. And interacting with such gizmo does math wrong. Translate, rotate and scale.
> 
>  2025-09-03.18-57-07.mp4

So here is what I have fixed
- Fixed descendant rotation (regression bug fix)
- The orthographic view I have followed blender and made the arc into a ring

> Is it possible to make a new gizmo look somewhat closer to existing ones (esp. the rectangle bit on translate)? So they don't look that "flat"?

Regarding the flatness I had made the gizmo opacity less so you can see the planes clearer however it was noted that the gizmos look less "vibrant". I can add further customisation to the gizmo theme in the future for separating all shapes however this is adequate for now.

Issue #1369, opened 2025-08-06
Clipboard (Copy/Paste) of field values (master ticket).


Maksims:
This ticket combines together the need to solve various ways of copying values of fields across the project's entities and assets.

Related tickets: #216 #30 #891 #520

### The desired UX:
1. It should be possible to copy/paste any modifiable value of entities and assets.
2. It should be clear on what fields support clipboard.
3. It should be clear on what field we are about to copy/paste when context menu is open.
4. The data should be validated and protected as much as possible to prevent undesirable data changes.
5. It should work with multi-selection (pasting) as well as copying (will copy first selected item's value).
6. History should have a single undo/redo operation for paste, even with multi-selection and multi-pathed fields (e.g.: particle curve sets; arrays of things)
7. User should see what type of value is about to be copied.
8. User should see what type of value is in the clipboard in context menu.

### Tasks:

- [ ] properly detect types of fields and their variations
- [ ] provide context menu on fields and their labels (except input and button dom elements)
- [ ] add copy/paste icons beside on fields hover - this will help to find out about feature and provide one-click workflow instead of context menu
- [ ] provide some sensible conversion paths between similar types, e.g. rgb <> rgba; asset:* > asset; asset > asset:* (if copied asset of desirable type); etc.
- [ ] support entity inspector
- [ ] support asset inspector
- [ ] use system's clipboard instead of localStorage by implementing value type detection from a JSON parsed string

### Screenshots:

<img width="342" alt="copy context menu" src="https://github.com/user-attachments/assets/9e2f8892-cfb0-479f-af46-8c10e1e3e453" />

Maksims:
Here is a video on current behavior and potential icons.

https://github.com/user-attachments/assets/8c333bd1-95e7-4aef-a1af-9774fef42ffb

Maksims:
Setting multiple entities from hierarchy to a field is now a breeze:

https://github.com/user-attachments/assets/2444a3cf-cd03-4622-9930-25c041e51cbb

willeastcott:
Can we close this yet @Maksims? Or do you intend to revisit any outstanding features related to this? One option is we close this and open individual issues for whatever's left, so it's clearer.

Maksims:
Can be added as individual tasks, yes.

willeastcott:
> Can be added as individual tasks, yes.

OK, closing. I've leave it to you to open any issues you still think need addressing. Thanks so much for all the hard work on this feature - it's awesome! 🚀

Issue #891, opened 2022-09-27
Allow copy-pasting fields of similar types


LeXXik:
It would be nice to be able to copy-paste an entity or a model reference from one component to another, instead of searching those in the hierarchies.

If the components are on the same entity, simply drag and drop to copy would suffice (or even drag and hover). If the components are on different entities, perhaps a copy/paste button could be added to the field.

![image](https://user-images.githubusercontent.com/5677782/192476391-630330be-02ec-4ef8-a466-64fb888803b8.png)

willeastcott:
@jiajasper I was thinking a right-click context menu with Copy + Paste on the attribute labels would be an option here. Do you have an alternative suggestions here?

yaustar:
Related tickets: https://github.com/playcanvas/editor/issues/520, https://github.com/playcanvas/editor/issues/216

Maksims:
> @jiajasper I was thinking a right-click context menu with Copy + Paste on the attribute labels would be an option here. Do you have any alternative suggestions here?

I was thinking on the same lines. As it allows to copy a complex fields e.g. vectors.

Maksims:
We really want this, as it becomes more and more common for a need to copy things within larger projects.

Even the small (gradual) implementation, like to start with: strings, number, vec2, vec3, vec4 - would be a massive addition.

Issue #520, opened 2021-08-31
Be able to copy and paste tags from one entity to another


yaustar:
Related issues:

https://github.com/playcanvas/editor/issues/216
https://github.com/playcanvas/editor/issues/204

Issue #30, opened 2020-03-12
Feature request: Copy entity position/rotate/size


qWici:
### Description
I think will be good if I can to copy some of entity properties like size, rotate or position.
As UX example: when right click on entity show:
```
Copy -> Entity
     -> Size
     -> Rotation
     -> Position
```

PR #1370, opened 2025-08-06
Clipboard - ability to copy/paste field values in Inspector

### LINKED ISSUE #216: Be able to copy component values from one Entity to another
Related to https://github.com/playcanvas/editor/issues/30

It would make iteration of entities a lot quickly, especially around the transforms and UI elements.

Maksims:
Fixes #216 #30 #891 #520

Master ticket: #1369

This PR introduces new feature for copy/paste values between various fields.
As fields have a value type (e.g. vec3, asset[], rgba, etc) we can store such info with the value in editor's clipboard (localStorage), and provide pasting capabilities of compatible types.

This is a first iteration, it only supports clipboard when entities are selected. Assets will come a bit later.

image

### DIFF
--- a/sass/editor/_editor-clipboard.scss
+++ b/sass/editor/_editor-clipboard.scss
@@ -1,0 +1,16 @@
+.pcui-contextmenu-clipboard {
+    > .pcui-menu-items {
+        > .pcui-menu-item {
+            > .pcui-menu-item-content {
+                > .pcui-menu-item-postfix {
+                    flex-grow: 1;
+                    text-align: right;
+
+                    &.disabled {
+                        display: none;
+                    }
+                }
+            }
+        }
+    }
+}
--- a/sass/editor/_editor-clipboard.scss
+++ b/sass/editor/_editor-clipboard.scss
@@ -1,3 +1,5 @@
+$highlight-color: #07f;
+
 .pcui-contextmenu-clipboard {
     > .pcui-menu-items {
         > .pcui-menu-item {
@@ -19,7 +21,7 @@
     content: '\00a0';
     position: absolute;
     inset: -4px;
-    border: 1px dotted #07f;
+    border: 1px dotted $highlight-color;
     z-index: 1;
     pointer-events: none;
 }
@@ -28,15 +30,15 @@
     content: '\00a0';
     position: absolute;
     inset: -4px;
-    border: 1px dotted #07f;
+    border: 1px dotted $highlight-color;
     z-index: 1;
     pointer-events: none;
     animation: pcui-clipboard-flash-animation 200ms ease-in-out forwards;
 }
 
 @keyframes pcui-clipboard-flash-animation {
-    from { border-color: #07f; }
-    to { border-color: rgba(0 128 255 0%); }
+    from { border-color: $highlight-color; }
+    to { border-color: rgba($highlight-color, 0.0); }
 }
 
 .pcui-label-group,
--- a/sass/editor/_editor-clipboard.scss
+++ b/sass/editor/_editor-clipboard.scss
@@ -23,3 +23,81 @@
     z-index: 1;
     pointer-events: none;
 }
+
+.pcui-element.pcui-highlight-flash::before {
+    content: '\00a0';
+    position: absolute;
+    inset: -4px;
+    border: 1px dotted #07f;
+    z-index: 1;
+    pointer-events: none;
+    animation: pcui-clipboard-flash-animation 200ms ease-in-out forwards;
+}
+
+@keyframes pcui-clipboard-flash-animation {
+    from { border-color: #07f; }
+    to { border-color: rgba(0 128 255 0%); }
+}
+
+.pcui-label-group,
+.pcui-asset-input {
+    &:hover {
+        > .pcui-label > .pcui-clipboard-button {
+            display: block;
+
+            &.pcui-hidden {
+                display: none;
+            }
+        }
+    }
+}
+
+.pcui-label {
+    &:hover {
+        > .pcui-clipboard-button {
+            display: block;
+
+            &.pcui-hidden {
+                display: none;
+            }
+        }
+    }
+
+    > .pcui-clipboard-button {
+        display: none;
+        position: absolute;
+        top: 0;
+        right: 0;
+        width: 16px;
+        height: 16px;
+        line-height: 16px;
+        margin: 0;
+        padding: 0;
+        text-align: center;
+        border: 0;
+        background-color: transparent;
+
+        &::before {
+            display: inline-block;
+            line-height: 16px;
+        }
+
+        &:hover,
+        &:focus {
+            box-shadow: none;
+            background-color: transparent;
+        }
+
+        &.pcui-clipboard-button-copy {
+            right: 16px;
+        }
+
+        &.pcui-hidden {
+            display: none;
+        }
+    }
+}
+
+.ui-tooltip > .inner.pcui-tooltip-disabled {
+    color: #777;
+}
--- a/sass/editor/_editor-clipboard.scss
+++ b/sass/editor/_editor-clipboard.scss
@@ -17,7 +17,7 @@ $highlight-color: #07f;
     }
 }
 
-.pcui-element.pcui-highlight::before {
+%pcui-highlight {
     content: '\00a0';
     position: absolute;
     inset: -4px;
@@ -26,13 +26,13 @@ $highlight-color: #07f;
     pointer-events: none;
 }
 
+.pcui-element.pcui-highlight::before {
+    @extend %pcui-highlight;
+}
+
 .pcui-element.pcui-highlight-flash::before {
-    content: '\00a0';
-    position: absolute;
-    inset: -4px;
-    border: 1px dotted $highlight-color;
-    z-index: 1;
-    pointer-events: none;
+    @extend %pcui-highlight;
+
     animation: pcui-clipboard-flash-animation 200ms ease-in-out forwards;
 }
 
--- a/sass/editor/_editor-clipboard.scss
+++ b/sass/editor/_editor-clipboard.scss
@@ -14,3 +14,12 @@
         }
     }
 }
+
+.pcui-element.pcui-highlight::before {
+    content: '\00a0';
+    position: absolute;
+    inset: -4px;
+    border: 1px dotted #07f;
+    z-index: 1;
+    pointer-events: none;
+}
--- a/src/editor/inspector/attributes-inspector.ts
+++ b/src/editor/inspector/attributes-inspector.ts
@@ -129,34 +129,63 @@ class AttributesInspector extends Container {
             evtChange = null;
         });
 
+        // if clipboard types are available
         if (this._clipboardTypes) {
+            // check if attribute has type and path and is of known type
             if (attr.type && attr.path && this._clipboardTypes.has(attr.type)) {
+                // once the field is parented
                 field.once('parent', (parent) => {
-                    if (!(parent instanceof LabelGroup)) {
-                        console.log(field);
-                        return;
-                    }
+                    let target = field;
 
-                    parent.dom.addEventListener('contextmenu', (evt) => {
-                        if (evt.target !== parent.label?.dom)
+                    // if part of a label group, provide copying for the whole element
+                    if (parent instanceof LabelGroup) target = parent;
+
+                    // try to bring context menu
+                    const onContextMenu = (evt) => {
+                        // do not interfere with inputs and buttons
+                        if (evt.target.tagName === 'BUTTON' ||
+                            evt.target.tagName === 'INPUT') {
+
                             return;
+                        }
 
+                        // supress default context menu
                         evt.stopPropagation();
                         evt.preventDefault();
 
-                        editor.call('clipboard:contextmenu:open', evt.clientX + 1, evt.clientY, attr.path);
+                        // modify type based on various rules
+                        let type = attr.type;
+                        if (type === 'select') type = attr.args.type;
+                        if (type === 'tags') type = `array:${attr.args.type}`;
+                        if (type === 'assets') type = 'array:asset';
+                        if (type === 'slider') type = 'number';
+                        if ((type === 'asset' || type === 'array:asset') && attr.args?.assetType) type += `:${attr.args.assetType}`;
+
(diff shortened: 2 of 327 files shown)


src/editor/inspector/attributes-inspector.ts
```diff
@@ -1,7 +1,9 @@
-import { Element, Container, LabelGroup, Panel, ArrayInput, BindingTwoWay } from '@playcanvas/pcui';
+import { Element, Container, LabelGroup, Panel, Button, ArrayInput, BindingTwoWay } from '@playcanvas/pcui';
 
 import { AssetInput } from '../../common/pcui/element/element-asset-input.ts';
 import { tooltip, tooltipRefItem } from '../../common/tooltips.ts';
+import { LegacyTooltip } from '../../common/ui/tooltip.ts';
```
kpal81xd, line 5: Please do not use LegacyTooltips we are trying to migrate away from Legacy UI to PCUI

willeastcott, line 5: Hey @kpal81xd - should this be using your new tooltip system?

src/editor/inspector/attributes-inspector.ts
```diff
@@ -126,6 +130,134 @@ class AttributesInspector extends Container {
...
+
+                    // if part of a label group, provide copying for the whole element
+                    if (parent instanceof LabelGroup) target = parent;
+
+                    // modify type based on various rules
+                    let type = attr.type;
+                    if (type === 'select') type = attr.args.type;
+                    if (type === 'assets') type = 'array:asset';
+                    if (type === 'slider') type = 'number';
+                    if ((type === 'asset' || type === 'array:asset') && attr.args?.assetType) type += `:${attr.args.assetType}`;
```
kpal81xd: Please use parentheses with these and/or using switch statements where necessary

src/editor/inspector/attributes-inspector.ts
```diff
@@ -126,6 +130,134 @@ class AttributesInspector extends Container {
...
+
+                    if (((target instanceof LabelGroup) || (target instanceof AssetInput)) && type !== 'label') {
+                        target.label.dom.style.position = 'relative';
+
+                        // paste button
+                        const btnPaste = new Button({
+                            icon: 'E353',
+                            class: 'pcui-clipboard-button'
+                        });
+                        btnPaste.enabled = false;
```
willeastcott: Moved into Button constructor args.

src/editor/inspector/attributes-inspector.ts
```diff
@@ -126,6 +130,134 @@ class AttributesInspector extends Container {
...
+                            if (pasted) editor.call('clipboard:flashElement', target.dom);
+                        });
+
+                        // copy button
+                        const btnCopy = new Button({
+                            // icon: 'E351',
+                            icon: 'E126',
+                            class: ['pcui-clipboard-button', 'pcui-clipboard-button-copy']
+                        });
+                        btnCopy.enabled = false;
```
willeastcott: Moved into Button constructor args.

sass/editor/_editor-clipboard.scss
```diff
@@ -0,0 +1,105 @@
...
+
+.pcui-element.pcui-highlight-flash::before {
+    content: '\00a0';
+    position: absolute;
+    inset: -4px;
+    border: 1px dotted $highlight-color;
+    z-index: 1;
+    pointer-events: none;
+    animation: pcui-clipboard-flash-animation 200ms ease-in-out forwards;
+}
```
kpal81xd: You can combine these by scoping the `.pcui-element.pcui-highlight` and by using `&::before` and `&::after`

src/editor/inspector/attributes-inspector.ts
```diff
@@ -47,6 +49,8 @@ class AttributesInspector extends Container {
         this._suspendChangeEvt = false;
         this._onAttributeChangeHandler = this._onAttributeChange.bind(this);
 
+        this._clipboardTypes = editor.call('clipboard:types') ?? null;
```
kpal81xd, line 54: Type editor call return values since type information is lost in event method

src/editor/storage/clipboard-context-menu.ts
```diff
@@ -0,0 +1,350 @@
...
+    let path: string | null = null;
+    let schemaType: string | null = null;
+    let elementHighlighted = null;
+
+    if (!clipboard) return;
+
+    // types of selected objects currently supported
+    const objTypes = new Set([
+        'entity'
+        // 'asset'
```
willeastcott: Maybe just add comment here such as "TODO: add support for assets"

src/editor/storage/clipboard-context-menu.ts
```diff
@@ -0,0 +1,350 @@
...
+    // context menu
+    const menu = new Menu({
+        class: 'pcui-contextmenu-clipboard'
+    });
+
+    // copy
+    const menuItemCopy = new MenuItem({
+        text: 'Copy',
+        icon: 'E351',
+        onSelect: function () {
```
willeastcott: Use arrow function.

src/editor/storage/clipboard-context-menu.ts
```diff
@@ -0,0 +1,350 @@
...
+            editor.call('clipboard:copy', path, schemaType);
+        }
+    });
+
+    // copy posftix that shows the type to be copied
+    const menuItemCopyLabel = new Label({
+        class: 'pcui-menu-item-postfix',
+        text: ''
+    });
+    menuItemCopy._containerContent.append(menuItemCopyLabel);
```
kpal81xd, line 116: Why are you accessing private properties?

src/editor/storage/clipboard-context-menu.ts
```diff
@@ -0,0 +1,350 @@
...
+        return true;
+    });
+
+    // flash dom element when copied/pasted
+    editor.method('clipboard:flashElement', (domElement: Element) => {
+        domElement.classList.add('pcui-highlight-flash');
+        setTimeout(() => {
+            domElement.classList.remove('pcui-highlight-flash');
+        }, 250);
+    });
```
willeastcott, line 386: PCUI has a function [`Element#flash`](https://api.playcanvas.com/pcui/classes/Element.html#flash) but I guess some of the UI is still legacy UI....is that why you're having to do this?

src/editor/storage/clipboard-context-menu.ts
```diff
@@ -0,0 +1,350 @@
...
+            menuItemPasteLabel.enabled = false;
+        }
+
+        // show context menu
+        menu.hidden = false;
+        menu.position(x + 1, y);
+    });
+
+    // copy to clipoard value by path from current selection
+    editor.method('clipboard:copy', (path, type) => {
```
kpal81xd: Add types to parameters

willeastcott (approved):
Approving with some minor comments/feedback. Nice one!! 🚀

src/editor/inspector/attributes-inspector.ts
Maksims: I've looked at the new tooltip system, and unfortunately it is not as good:
1. It has a delay where it is undesirable.
2. It is not well aligned
3. The styling of it is not as clear.

src/editor/storage/clipboard-context-menu.ts
Maksims: They are different. Element#flash uses outline on element. While it is ok on many elements like button or input, it does not look good on LabelGroup element due to lack of padding. So pcui-highlight-flash class implements highlight with :before class with padding to better outline whole field.

src/editor/inspector/attributes-inspector.ts
Maksims: While the currently used tooltip works flawlessly.
The new one simply did not work, and it has delay where it does not need one.
image

src/editor/inspector/attributes-inspector.ts
Maksims: Switch wont work here, as some rules apply to both cases: if no rules before applied as well as some rules before applied.
It is easy to follow the logic of rules here.

src/editor/storage/clipboard-context-menu.ts
willeastcott: Looks like it's because the container isn't accessible via the PCUI API at the moment.

Maksims: Because I need value of that property.

willeastcott: > Because I need value of that property.

I think my answer was better. 😉

src/editor/inspector/attributes-inspector.ts
kpal81xd: That is because its being assigned to the same base tooltip. Its fine I will migrate all the legacy tooltips as required

src/editor/storage/clipboard-context-menu.ts
kpal81xd: Okay add a comment to explain that its not exposed

src/editor/inspector/attributes-inspector.ts
```diff
@@ -126,6 +132,133 @@ class AttributesInspector extends Container {
...
+
+                                const clipboardType = editor.call('clipboard:typeToHuman', editor.call('clipboard:type'));
+                                tooltipPaste.text = `Paste${clipboardType ? ` ${clipboardType}` : ''}`;
+                                tooltipPaste.innerElement.classList.toggle('pcui-tooltip-disabled', !btnPaste.enabled);
+                            }
+                        });
+                    }
+
+                    // TODO
+                    // extra rule for array of assets, to copy individual assets within the group
```
kpal81xd:
```suggestion
                    // TODO:
                    // extra rule for array of assets, to copy individual assets within the group
```

src/editor/inspector/attributes-inspector.ts
```diff
@@ -126,6 +132,133 @@ class AttributesInspector extends Container {
...
+                });
+            } else if (attr.type && attr.paths) {
+                // TODO
+                // implement copy-paste for multi-path fields
```
kpal81xd:
```suggestion
                // TODO:
                // implement copy-paste for multi-path fields
```

src/editor/storage/clipboard-context-menu.ts
```diff
@@ -0,0 +1,352 @@
...
+
+    if (!clipboard) return;
+
+    // types of selected objects currently supported
+    const objTypes = new Set([
+        'entity'
+    ]);
+
+    // TODO
+    // add support for 'asset' inspector
```
kpal81xd:
```suggestion
    // TODO:
    // add support for 'asset' inspector
```

src/editor/storage/clipboard-context-menu.ts
```diff
@@ -0,0 +1,352 @@
...
+            value.hasOwnProperty('value') &&
+            value.type;
+    };
+
+
+    // TODO
+    // type to other type conversions, e.g.:
+    // asset:* > asset
+    // asset > asset:* - if copied asset is of desired type
+    // rgb <> rgba
```
kpal81xd:
```suggestion
    // TODO:
    // type to other type conversions, e.g.:
    // asset:* > asset
    // asset > asset:* - if copied asset is of desired type
    // rgb <> rgba
```

src/editor/storage/clipboard-context-menu.ts
```diff
@@ -0,0 +1,352 @@
...
+        }
+
+        // should have at least one item in selector
+        const items = editor.call('selector:items');
+        if (!items.length) return false;
+
+        const paste = clipboard.value;
+
+        // TODO
+        // verify if value is actually valid based on type
```
kpal81xd:
```suggestion
        // TODO:
        // verify if value is actually valid based on type
```

src/editor/storage/clipboard-context-menu.ts
```diff
@@ -0,0 +1,352 @@
...
+                valueNew: paste.value
+            });
+
+            // paste new value
+            items[i].history.enabled = false;
+            items[i].set(path, paste.value);
+            items[i].history.enabled = true;
+
+            // TODO
+            // setting render-component asset does not update materials - bug in render component inspector
```
kpal81xd:
```suggestion
            // TODO:
            // setting render-component asset does not update materials - bug in render component inspector
```

src/editor/inspector/attributes-inspector.ts
Maksims: Not sure I actually understand this comment.
I've added above the field: `_clipboardTypes: Set<string> | null;`, which helps to satisfy this?

Maksims:
TODOs will be added as subsequent PR's to keep this PR more focused. As it is already useful to the users.

Maksims:
The newly introduced rule on linter:
```
326:32  error  Expected { after 'if' condition  curly
```
Making readability ugly.

In many cases of simple "return early", now instead of simply:

```js
if (!item) continue;
```

It becomes:
```js
if (!item) {
    continue;
}
```

You guys sure about that?

Specifically on this PR, this readable states:

```js
let type = attr.type;
if (type === 'select') type = attr.args.type;
if (type === 'assets') type = 'array:asset';
if (type === 'slider') type = 'number';
if ((type === 'asset' || type === 'array:asset') && attr.args?.assetType) type += `:${attr.args.assetType}`;
```

Became this:
```js
if (type === 'select') {
    type = attr.args.type;
}
if (type === 'assets') {
    type = 'array:asset';
}
if (type === 'slider') {
    type = 'number';
}
if ((type === 'asset' || type === 'array:asset') && attr.args?.assetType) {
    type += `:${attr.args.assetType}`;
}
```

Maybe lets focus on doing features, instead of styling and moving the code around, please?

kpal81xd:
That rule change was less 5 seconds to implement and even less time/effort to apply automatically so I do not think it takes time away from implementing features

kpal81xd (approved):

willeastcott:
OK, guys - we're good to merge now? 😄

kpal81xd:
> OK, guys - we're good to merge now? 😄

Yep all good from me 🫡

willeastcott:
Let's gooOOOooooOOO!!! Merging. Thanks @Maksims!

Maksims:
Thank you for the review!

PR #1375, opened 2025-08-07
Refactor JSDoc handling in attribute autofill logic

marklundin:
This PR significantly refactors the user flow to promote attributes and class members are handled in the Code editor.

Before:

Clipboard-20250807-142557-940

- UX to promote a class member is via nested quick fix
- Remove attribute destroys all JSDoc tags
- Quick fix displays arbitrary json

After:

Clipboard-20250807-142415-921

- UX to promote a member is up-front and clear
- Remove attribute only destroys attribute tags, not other tags ie @description or @title
- No arbitrary json displayed

## Summary

* Introduced JSDocUtils class for managing JSDoc operations, including analysis, creation, and cleaning of JSDoc blocks.
* Enhanced modifyJSDocAttribute function to handle adding and removing JSDoc attributes more effectively.
* Updated code lens provider to support new JSDoc actions for attributes, including adding slider functionality.
* Improved error handling in the ESM script worker by clarifying error property names in documentation.

### DIFF
--- a/src/code-editor/monaco/intellisense/attribute-autofill.ts
+++ b/src/code-editor/monaco/intellisense/attribute-autofill.ts
@@ -22,19 +22,17 @@ import { WorkerClient } from '../../../core/worker/worker-client.ts';
 /**
  * @typedef {object} ParseAttribute
  * @property {boolean} isAttribute - Whether the attribute is an attribute
- * @property {object} start - The start position
+ * @property {string} name - The member name
+ * @property {string} type - The attribute type (e.g., 'number', 'string')
+ * @property {object} start - The start position (JSDoc start if exists, otherwise member position)
  * @property {number} start.line - The start line number
  * @property {number} start.column - The start column number
- * @property {object} end - The end position
+ * @property {object} end - The end position (JSDoc end if exists, otherwise member position)
  * @property {number} end.line - The end line number
  * @property {number} end.column - The end column number
  */
 
 const OWNER = 'attribute-autofill';
-const CHAR_LIMIT = 1e6;
-const ERROR_LIMIT = 1000;
-const ATTRIBUTE_LIMIT = 1000;
-const modelMarkersWithFixes = new Map();
 
 /**
  * @param {string} url - The engine URL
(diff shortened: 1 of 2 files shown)


Maksims:
Jumping lines feels awkward.

Please don't change line height or append anything above that leads to jumping lines. There should be a better way to provide some clickable buttons without interfering with lines and their consistent heights.

marklundin:
@Maksims if you want to add an attribute, you have to add a `/** @attribute */` tag, so not sure what can be done here. 

If a line is added in response to a user action, I think it's fine. Code Lenses are  standardized UX in vscode https://code.visualstudio.com/blogs/2017/02/12/code-lens-roundup

Maksims:
I'm refering to "Add attribute" and "Remove attribute" buttons, they appear and disappear franticly, and when that happens whole code below is jumping up and down.
Can that be avoided please?

marklundin:
Yep I think we can avoid that flicker as it adds and removes the code lens. This should avoid that visual jump between states

kpal81xd (changes requested):
Use typescript types instead of JSDoc pls

marklundin:
@Maksims This has been updated so things don't jump around. The actions just persist and update as the model changes. Makes things a little more visually consistent 😅 

@kpal81xd Updated to TS Types

src/code-editor/monaco/intellisense/attribute-autofill.ts
```diff
@@ -1,87 +1,209 @@
...
- * @property {number} end.line - The end line number
- * @property {number} end.column - The end column number
- */
+interface ParseAttribute {
+    isAttribute: boolean;
+    name: string;
+    type: string;
+    start: { line: number; column: number };
+    end: { line: number; column: number };
+}
```
kpal81xd:
```suggestion
type ParseAttribute = {
    isAttribute: boolean;
    name: string;
    type: string;
    start: { line: number; column: number };
    end: { line: number; column: number };
}
```

Types are preferable over interfaces

src/code-editor/monaco/intellisense/attribute-autofill.ts
```diff
@@ -1,87 +1,209 @@
...
+     * @param {ParseAttribute} attribute - The attribute object with JSDoc positioning
+     * @returns {{ hasJSDoc: boolean, isSingleLine: boolean }} - The JSDoc analysis result
+     */
+    static analyzeJSDoc(attribute) {
+        // If start == end (same position), no JSDoc exists
+        const hasJSDoc = !(attribute.start.line === attribute.end.line && attribute.start.column === attribute.end.column);
+        // If JSDoc exists and spans only one line, it's single-line
+        const isSingleLine = hasJSDoc && attribute.start.line === attribute.end.line - 1;
+        return { hasJSDoc, isSingleLine };
+    }
```
kpal81xd: Use TSDoc and fold types into arguments

src/code-editor/monaco/intellisense/attribute-autofill.ts
```diff
@@ -1,87 +1,209 @@
...
-    const scripts = await Promise.all(assets.reduce((/** @type {Promise<[string, string]>[]} */ acc, asset) => {
+    const scripts = await Promise.all(assets.reduce((acc: Promise<[string, string]>[], asset: Observer) => {
         // skip non-module assets
         if (!editor.call('assets:isModule', asset)) {
             return acc;
         }
 
         // skip filtered paths
-        const path = /** @type {string} */ (editor.call('assets:virtualPath', asset));
+        const path = editor.call('assets:virtualPath', asset) as string;
```
kpal81xd:
```suggestion
        const path: string = editor.call('assets:virtualPath', asset);
```

Try not to use `as <type>` where possible

src/code-editor/monaco/intellisense/attribute-autofill.ts
```diff
@@ -90,298 +212,204 @@ const fetchModuleScripts = async (cache, filter = []) => {
...
- *
- * @param {Record<string, string>} tags - Object with attribute properties
- * @returns {string} JSDoc attribute
+ * Generic function to modify JSDoc attributes
+ * @param {Monaco.editor.ITextModel} model - The text model
+ * @param {number} lineNumber - The line number of the member
+ * @param {string} memberName - The name of the member
+ * @param {ParseAttribute} attribute - The attribute object with JSDoc positioning
+ * @param {string} action - The action to perform
  */
```
kpal81xd: Can remove the type here
```suggestion
 * Generic function to modify JSDoc attributes
 * @param model - The text model
 * @param lineNumber - The line number of the member
 * @param memberName - The name of the member
 * @param attribute - The attribute object with JSDoc positioning
 * @param action - The action to perform
 */
```

Can remove the types to use TSDoc or can omit the tsdoc entirely its clear enough through the type and parameter

marklundin: We have `eslint[jsdoc/require-param](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/require-param.md#repos-sticky-header)` rules, so until that's removed it will fail linting @kpal81xd

src/code-editor/monaco/intellisense/attribute-autofill.ts
```diff
@@ -1,87 +1,196 @@
...
+) : Promise<[[string, string][], string[]]> => {
+    const assets = (editor.call('assets:list') ?? []) as Observer[];
 
     // Get all the files that no longer exist. ie files in the cache, but not in esmScripts
     const deletedFiles = [];
 
     // loop over the file cache, remove any files that do no exist in the script assets
-    const assetPaths = assets.map(asset => editor.call('assets:virtualPath', asset));
-    for (const filePath in cache) {
+    const assetPaths = assets.map(asset => editor.call('assets:virtualPath', asset) as string);
```
kpal81xd:
```suggestion
    const assetPaths: string[] = assets.map(asset => editor.call('assets:virtualPath', asset));
```

src/code-editor/monaco/intellisense/attribute-autofill.ts
```diff
@@ -1,87 +1,196 @@
...
         // skip filtered paths
-        const path = /** @type {string} */ (editor.call('assets:virtualPath', asset));
+        const path: string = editor.call('assets:virtualPath', asset);
         if (filter.includes(path)) {
             return acc;
         }
 
         // skip cached files
-        const hash = asset.get('file.hash');
+        const hash = asset.get('file.hash') as string;
```
kpal81xd:
```suggestion
        const hash: string = asset.get('file.hash');
```

src/code-editor/monaco/intellisense/attribute-autofill.ts
```diff
@@ -471,17 +496,199 @@ editor.once('load', () => {
...
+            provideCodeActions: (model) => {
+
+                const actions = [];
+
+                if (!busy) {
+                    // Get all markers for this model, not just context markers
+                    const allMarkers = monaco.editor.getModelMarkers({ owner: OWNER });
+                    const modelMarkers = allMarkers.filter(marker => marker.resource.toString() === model.uri.toString());
+
+                    modelMarkers.forEach((/** @type {Monaco.editor.IMarker} */ marker) => {
```
kpal81xd:
```suggestion
                    modelMarkers.forEach((marker: Monaco.editor.IMarker) => {
```

PR #1395, opened 2025-08-13
Clipboard to respect enum/select type fields in Inspector

Maksims:
Fixes task in #1369

Select type fields can have a type of value like string or number, and then options to select from. While clipboard simply cares about type, this PR introduces a validation step, of ensuring that we can only paste a value if it is one of select options.


kpal81xd (approved):
Nice!

PR #1404, opened 2025-08-20
PCUI update to 5.2.0

kpal81xd:
- Updates PCUI to 5.2.0
- Fixed import of theme in error sass file
- Default Root entity in treeview to be open

### DIFF
--- a/src/editor/entities/entities-treeview.ts
+++ b/src/editor/entities/entities-treeview.ts
@@ -598,7 +598,7 @@ class EntitiesTreeView extends TreeView {
                 dropType === 'asset.sprite';
     }
 
-    _onAddEntity(entity) {
+    _onAddEntity(entity, open: boolean = false) {
         const resourceId = entity.get('resource_id');
         if (this._treeItemIndex[resourceId]) {
             return this._treeItemIndex[resourceId];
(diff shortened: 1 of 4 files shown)


willeastcott (approved):
Maybe just verify that both the asset and entity tree views default to open.

src/editor/entities/entities-treeview.ts
```diff
@@ -957,7 +958,7 @@ class EntitiesTreeView extends TreeView {
         if (this._entities) {
             this._eventsObserverList.push(this._entities.on('add', this._onAddEntity.bind(this)));
             this._eventsObserverList.push(this._entities.on('remove', this._onRemoveEntity.bind(this)));
-            this._entities.forEach(entity => this._onAddEntity(entity));
+            this._entities.forEach((entity, i) => this._onAddEntity(entity, i === 0));
```
willeastcott: Maybe consider just constructing the tree as normal and just setting `open` to `true` on the root `TreeViewItem` afterwards.

willeastcott: Hmm! Interesting. You can't easily access the root(s) of a `TreeView`. Looks like you'd have to do something like:

```ts
function getRoot(treeView: TreeView): TreeViewItem | null {
    let rootItem: TreeViewItem | null = null;
    
    treeView.forEachChild((child: Element) => {
        if (child instanceof TreeViewItem) {
            rootItem = child;
            return false; // Early out after finding the first TreeViewItem
        }
    });
    
    return rootItem;
}
```

...which assumes the `TreeView` has just one root (which is does in the Asset and Entity trees).

TBH, your solution is more succinct so maybe stick with that. We could probably improve the PCUI API in this area.

kpal81xd: There is a check for the parent to acquire `rootItem` in`_onAddEntity` so I am just using that

Issue #1159, opened 2024-05-21
Do not allow creating a script with a reserved name


LeXXik:
At the moment you can create a script with any name, even with restricted name. Then the Editor will show an error (an obscure one, at the bottom status line) that the script is using a reserved name. This is a feature request to not allow creating a script with a reserved name in the first place.

![image](https://github.com/playcanvas/editor/assets/5677782/af588920-13cf-4750-98ca-fad801a489d1)

mvaligursky:
related / duplicate of https://github.com/playcanvas/editor/issues/927

LeXXik:
Closing in favor #927

PR #1424, opened 2025-09-04
Validate against reserved script names when creating a script

### LINKED ISSUE #927: When creating a script, validate its name against the reserved names list
see details on getReservedScriptNames : https://github.com/playcanvas/engine/pull/4847

abstrakt8:
### Issues

* Fixes #927 
* Fixes (better description): #1159 

### What was changed

* Created a reusable function `normalizeScriptName`: 
  * Trims whitespaces
  - Appends a `.js` extension if the current one is not `.js` or `.mjs`.
  - Does some basic validation checks equivalent to the existing ones.
  - (**New**) Checks against the reserved names by using `getReservedScriptNames` exported from `playcanvas`. See [here](https://github.com/playcanvas/engine/blob/0b428c9a654fd301ec2a635bcc62e66895f1bd46/src/framework/script/script-create.js#L8-L16).
  - See unit tests for more info

* Included the `normalizeScriptName` in the Editor's and CodeEditor's script creator floating modals to prevent the creation of scripts with invalid names. 

### Testing 

Using the reserved keyword `data`:

Screenshot 2025-09-04 at 14 35 20

Screenshot 2025-09-04 at 14 35 53

---

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -104,6 +104,7 @@ export default [
         rules: {
             'accessor-pairs': ['error', { setWithoutGet: false, getWithoutSet: false }],
             'curly': ['error', 'all'],
+            'no-unused-expressions': 'off',
             'jsdoc/require-param-type': 'off',
             'jsdoc/require-returns-type': 'off',
             'jsdoc/require-returns': 'off',
--- a/test/common/script-names.test.ts
+++ b/test/common/script-names.test.ts
@@ -1,7 +1,8 @@
 import { expect } from 'chai';
 import { describe, it } from 'mocha';
-import { normalizeScriptName } from '../../src/common/script-names.ts';
 import { getReservedScriptNames } from 'playcanvas';
+
+import { normalizeScriptName } from '../../src/common/script-names.ts';
 
 describe('normalizeScriptName', () => {
     it('should return the original when it is valid', () => {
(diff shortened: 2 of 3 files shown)


src/common/script-names.ts
```diff
@@ -0,0 +1,43 @@
+import { getReservedScriptNames } from 'playcanvas';
+
+// Should not start with a number and then should not contain invalid chars
+const sanityRegex = /^[^0-9.#<>$+%!`&='{}@\\/:*?"|\n][^#<>$+%!`&='{}@\\/:*?"|\n]*$/;
```
abstrakt8, line 4: Not quite sure behind the origin of this rule, but I just kept it as it is.

src/code-editor/pickers/picker-script-create.ts
```diff
@@ -1,8 +1,9 @@
 import { Overlay, Label, TextInput } from '@playcanvas/pcui';
 
+import { normalizeScriptName } from '../../common/script-names.ts';
```
kpal81xd, line 3: I would move this to the utils file no need to create an entirely new file for it

src/common/script-names.ts
kpal81xd: Check git blame in https://github.com/playcanvas/editor-ui

eslint.config.mjs
```diff
@@ -104,6 +104,7 @@ export default [
         rules: {
             'accessor-pairs': ['error', { setWithoutGet: false, getWithoutSet: false }],
             'curly': ['error', 'all'],
+            'no-unused-expressions': 'off',
```
abstrakt8, line 107: Same as in `engine`:

https://github.com/playcanvas/engine/blob/0b428c9a654fd301ec2a635bcc62e66895f1bd46/eslint.config.mjs#L58

So that expressions like `.to.be.null;` do not give any errors

src/code-editor/pickers/picker-script-create.ts
abstrakt8: Discussed offline to leave it for now. Something like a `utils` folder will be created and we can move it inside.

kpal81xd (approved):
Nice congrats on your first PR approval 🎉

abstrakt8:
Thanks @kpal81xd !

PR #1428, opened 2025-09-09
Gizmo overhaul for translate/rotate/scale

kpal81xd:
Continuation of #1367

### Requires
- https://github.com/playcanvas/engine/pull/7960

### What's changed
- Replaces translate/rotate/scale gizmos with ones from engine
- Added Gizmo customisation settings #1431 

### Testing
- Engine `v2.12.4`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@playcanvas/editor",
-  "version": "2.6.6",
+  "version": "2.6.7",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@playcanvas/editor",
-      "version": "2.6.6",
+      "version": "2.6.7",
       "license": "MIT",
       "dependencies": {
         "@playcanvas/attribute-parser": "^1.10.1"
@@ -16,7 +16,7 @@
         "@jsquash/jpeg": "1.4.0",
         "@jsquash/png": "2.2.0",
         "@jsquash/webp": "1.4.0",
-        "@playcanvas/editor-api": "^1.1.25",
+        "@playcanvas/editor-api": "^1.1.26",
         "@playcanvas/eslint-config": "2.1.0",
         "@playcanvas/observer": "1.6.6",
         "@playcanvas/pcui": "5.2.4",
@@ -601,9 +601,9 @@
       }
     },
     "node_modules/@playcanvas/editor-api": {
-      "version": "1.1.25",
-      "resolved": "https://registry.npmjs.org/@playcanvas/editor-api/-/editor-api-1.1.25.tgz",
-      "integrity": "sha512-wN150tsLD0jCDFqxrMN7vp/+0+N4iPuYYmsYHnEEY9hXWza74fW91pbPZVLzNB8PaxmLHX3cFs52mNrDHe5/OQ==",
+      "version": "1.1.26",
+      "resolved": "https://registry.npmjs.org/@playcanvas/editor-api/-/editor-api-1.1.26.tgz",
+      "integrity": "sha512-eymfzhmVfEpPj1zv107LfihR25QQwVY7svHRuyPYRkY7jTfyyDwscVB6kVqL3NfZ9eEEcLYz8ibcS+ERyeidkQ==",
       "dev": true,
       "license": "MIT",
       "engines": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@playcanvas/editor",
-  "version": "2.6.6",
+  "version": "2.6.7",
   "author": "PlayCanvas <support@playcanvas.com>",
   "homepage": "https://playcanvas.com",
   "description": "The PlayCanvas Editor Frontend",
@@ -32,7 +32,7 @@
     "@jsquash/jpeg": "1.4.0",
     "@jsquash/png": "2.2.0",
     "@jsquash/webp": "1.4.0",
-    "@playcanvas/editor-api": "^1.1.25",
+    "@playcanvas/editor-api": "^1.1.26",
     "@playcanvas/eslint-config": "2.1.0",
     "@playcanvas/observer": "1.6.6",
     "@playcanvas/pcui": "5.2.4",


willeastcott (approved):

Maksims:
**This PR changes visuals and most importantly mechanics of how gizmos behave.**

This is extremely undesirable, as users got intuition and comfort of using current gizmos, and just abruptly without any warning or choice forcing everyone to re-learn, without any benefits, with actual reduction in experience.

Very very undesirable.

## Issues:
1. Flipping axes arrows prevents us from knowing where forward and where back is, same with other axes.
2. Rotation two modes gizmo is magically switching between, and is an awful UX of hidden rules.
3. Visually gizmos are dull and non-clear, mixing up with our content, making it harder to interact with.

Maksims:
Please vote on this poll if these changes are desirable or will make you unhappy: https://x.com/mrmaxm/status/1966079197618847987

yaustar:
I'm not a huge fan of the axis flipping behaviour as a default setting as that's quite confusing. The other issues that Max has raised I personally don't see as deal breakers (colours could (and maybe should?)) be added as presets in settings but the axis flipping I personally find very confusing as I have to mentally flip the axis to know the orientation of the object when selected

https://github.com/user-attachments/assets/0eb97326-68ce-4874-9eb6-b04cb029e3fb

Edit: The more I use the new rotation gizmo, I think it's fine but likely to need a sensitivity setting 

Edit2: Ah sorry, knee jerk reaction. Looks like there is a setting to switch between the two https://github.com/playcanvas/editor/pull/1431

OlegGedzjuns:
https://github.com/user-attachments/assets/c1ff51da-e3dd-46fa-ae30-196207129821
How do i even know now where is forward/back and left/right of my entity with arrows always flipping???


chrome_kYJRXH0zeF
If i move entity right now, will my entity move like i drawn here? NOPE, i just rotated my camera around entity and coordinate changes will be exactly opposite

Yeah, I see that there is option to disable axes flipping, but this SHOULD NOT be default behaviour

LeXXik:
I usually use the arrows as an indication of which direction is a positive or negative axis in relation to the object. I can see it by selecting an entity. With flipping arrows, I would have to move the object to see how the values change in order to find out the same info. Not sure I like it, but not going to fuss about it either. A matter of habbit I guess. Scale and rotation are fine for me.

kpal81xd:
The axis flipping was added so that the gizmos would still be reachable if covered by other parts (i.e.. the planes). 

The line connected to the center was removed to show direction but seems like its not clear enough.

I am going to remove the axis flipping in favor of better positioning of planes (like how Blender and Unity handle it)

Maksims:
When flying with WASDQE, raycasting is not happening, only when mouse is moved, leading to "croppy" movement of the gizmos.

Video shows how it works currently (correct), and how it works with this PR (incorrect).

https://github.com/user-attachments/assets/cc58c6f8-5b23-42d7-8b28-3bd4754dc719

Maksims:
Regression:

It is not possible with a new Gizmo (this PR) to grab object, and use look around (right mouse button) while dragging the object.

Video below shows current behavior, that allows to drag object, and re-orient camera while doing so, while the new gizmo, simply ignores right click, leading to inability to re-orient, which forces to release first, re-orient, and then grab again. We use look around with WASD in our projects pretty much everywhere, and this will be a regression in UX.

https://github.com/user-attachments/assets/acb21ba5-7179-4449-91a8-91bae9ef7846

Maksims:
While new rotation gizmo visually is an upgrade, it's math is changed from the current one, and behaves inconsistently. Depending on steepness and where the circle is grabbed from, the rotation direction will not be towards the mouse, but will move to some offset for some reason.

Here is an image, with blue pointer - where mouse is, and red line - where the rotation vector is pointing (wrong), while it should be pointing as a green line (correct). Current gizmos are doing the correct math currently.

image



Video below shows current correct and clear behavior, then shows this PR gizmo with weird offset when rotating:

https://github.com/user-attachments/assets/c60c0c7d-7b69-4dc4-9f84-1badd3517a3a

Maksims:
Regression with multi-selection and local space gizmos.

Currently, when multi-selecting, and using local space, gizmo will be positioned at the last selected entity, and will be rotated based on that entity. So manipulating that gizmo, behaving same as manipulating a single entity, no funky weird stuff. Other entities will transform in their local space.

With this PR, the rotation of gizmo will be based on last selected entity, but the position will be based on center of selection's AABB - so when you interact with the gizmo, it does not follow mouse at all, and moves around in weird directions. Also when pointer up, the gizmo will not be where your mouse is, and be moved to another average place, with more complex selections, user has to "chase" the gizmo around. And won't be able to move objects where they expect.

Video shows, how it is now (correct), and how it is with this PR:

https://github.com/user-attachments/assets/321b0097-6011-40ad-babc-1f681fa5b463

Maksims:
On Classic gizmos:

## A few new features that feels **good**:
1. Rotation using view-aligned axis - feels great, with rotation motion of mouse around the object.
2. Visuals of rotation gizmo - is a good upgrade vs current one.
3. Axes guides changes - generally feels like an improvement.

## Some that feels as **downgrade** in addition to logical described in messages above:
1. Translation gizmo visually feels as a downgrade, lines are too thin, rectangle is not solid, feels too loud, and does not help to visually see axis arrows.
2. When we are axis-aligned with rotation gizmo, then "big circle" - is not needed, and could be hidden.


## On "default" gizmos (new):

1. Translation gizmos have arrows shown for some reason when moving, unnecessary obstruction from seeing what we are translating.
2. Colors are dull, and easy to mix with scene.
3. Hovering states feel too busy and convoluted, could be cleaner and more simple.
4. Why it is needed in the first place, except rotation gizmo rotation on non-axis, and translation on screen-plane - there is no improvements, or benefits of having different gizmos.

### Why not make current (classic) gizmo behavior slightly better, and simply ship it, without complicating code-base with two versions of gizmos, and also splitting users with two versions of gizmos? Why complicating it all?

kpal81xd:
@Maksims thanks again for the QA review 😄 

1. & 2.  For flying and rotating while holding the gizmo I have added (surprised this slipped though earlier)

3. The rotation was changed to match Blender to rotating slight angles becomes easier but this makes the guide angle obsolete and requires another guide to show rotation. Scrapped the blender rotation and reverted to the old rotation style

4. Local space gizmo position yep I agree it makes no sense to use average when spaces are different so it uses the last selection space for the gizmos transform

Maksims:
Let us use this gizmos in work for a day, so we can battle test it, and see if we are all good.

Maksims:
Minor one, but nonetheless:

When using ortho camera, the rectangle in translate gizmo, is flipping around - this is actually a current behavior also.
I'm wandering if it has to flip at all.

But other small bit, is when rectangle is flipped, the length of an arrow, is not extended to the middle of the gizmo:

Current:
image

New:
image

Notice that red (X) axis line, starts not from middle when rectangle is flipped, but from far away, making it non-hoverable from the middle.

Also the line thickness feels like less than 1 pixel, making it very thing compared to other lines.

Also rectangle style, I personally like the current one way more, than this solid one.

kpal81xd:
> Minor one, but nonetheless:
> 
> When using ortho camera, the rectangle in translate gizmo, is flipping around - this is actually a current behavior also. I'm wandering if it has to flip at all.
> 
> But other small bit, is when rectangle is flipped, the length of an arrow, is not extended to the middle of the gizmo:
> 
> Current: image
> 
> New: image
> 
> Notice that red (X) axis line, starts not from middle when rectangle is flipped, but from far away, making it non-hoverable from the middle.
> 
> Also the line thickness feels like less than 1 pixel, making it very thing compared to other lines.
> 
> Also rectangle style, I personally like the current one way more, than this solid one.

@Maksims thanks for testing 🙂 . I will probably do some follow up PRs in the engine down the road to address the inconsistencies in the gizmo design like the line thickness, length and the rectangle opacity but this PR I just want to make sure they function at least as good as the existing ones.

PR #1431, opened 2025-09-10
Gizmos settings

kpal81xd:
### Requires
- Backend schema update for project user settings
- https://github.com/playcanvas/engine/pull/7961

### Description
- Adds gizmo customisation settings under `Editor` settings header
  - Gizmo size project user setting. Defaults to `1.2`
  - Gizmo preset project user setting (`default` or `classic`). Defaults to `default`.

### Preview
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,7 +42,7 @@
         "monaco-themes": "^0.4.2",
         "nodemon": "3.1.7",
         "ot-text": "github:playcanvas/ot-text",
-        "playcanvas": "^2.11.3",
+        "playcanvas": "^2.11.4",
         "postcss": "8.4.48",
         "rollup": "4.25.0",
         "rollup-plugin-copy": "3.5.0",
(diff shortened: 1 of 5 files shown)


mvaligursky:
If you default to classic, nobody will see the improved version.

kpal81xd:
> If you default to classic, nobody will see the improved version.

I wanted to keep the behavior the same for existing users but yea otherwise ill have to do a migration so prob better to default to improved version

Maksims:
It should be "classic" by default for all existing users.
Users have muscle memory, any forced and "did not asked for" changes on their daily life - will be met with anger and frustration. Please do not change current behavior.

For all new registered users, new behavior can be applied to.

If users don't find this setting, perhaps they don't even asked for it then, and do not need it in the first place.
If forced on users, how do they know that there is a setting to go back to "classic"?

So please, don't change it for everyone! Keep default existing behavior for existing users. Let them do their work and be with their lives without need to get forcibly learning something they did not asked for, and will not allow them to make better content.

Maksims:
Based on even this basic poll.
You have a choice:
1. Make default for existing users to forcibly switch to new behavior. That will lead to 50%+ of unhappy users, damaging trust. And 50%- of potentially happy users.
2. Make default for existing users to stay as is. No unhappy users! Keep the trust. Allow some users to switch to new behavior - more happy users.

1st option - **negative** gain.
2nd option - only **positive** gain.

Simple.

image

PR #1435, opened 2025-09-11
Template entity inspector: Link to the newly created template, when inspecting an entity without a template

### LINKED ISSUE #1018: Templates missing names
When a new tamplate is created, it is missing a name in the properties:
image

If the template is deselected and selected again, the name is shown correctly.

### LINKED ISSUE #1392: Creating new template from existing one shows incorrect reference name
Steps to reproduce:

1. Initialize new exisitng template
2. Unlink from template
3. Change name
4. Create a new template
5. Template reference name is incorrect

https://github.com/user-attachments/assets/32284247-afcb-4ab6-a216-4d2af617f90c

### LINKED ISSUE #1417: After creating template, template reference has empty name
https://github.com/user-attachments/assets/53e5dab7-611d-425e-9a8a-c50237bef6d8

abstrakt8:
### Issue
Fixes #1392 

After reading through the code, I found another set of reproduction steps (to #1392):
* Select any entity A with a template 
* Select any other entity B **without** a template
* Create a new template from that entity and notice that the template name has that of A.

As a side effect, it also fixes the following:
Fixes #1417
Fixes #1018

### What has changed?

* After calling the `TemplateOverrideInspector#link`, we don't refresh the `templateAsset` in case of an `template_id:set` event. I moved the code of setting the `templateAsset` and its corresponding label to the `refreshOverrides` function
* The `refreshOverrides` function used to recalculate the overrides, but it also used to **only** detect whether a template has been unlinked and apply the corresponding changes (but not vice versa). 

### Testing

* Tested it with the reproduction steps in the issues above

### DIFF
--- a/src/editor/assets/assets-upload.ts
+++ b/src/editor/assets/assets-upload.ts
@@ -275,7 +275,8 @@ editor.once('load', () => {
                 data: data,
                 meta: asset ? asset[1].get('meta') : null
             }, (err, data) => {
-                if (err || ext !== 'js') {
+                const isJavascript = ext === 'js' || ext === 'mjs';
+                if (err || !isJavascript) {
                     return;
                 }
 


abstrakt8:
https://github.com/user-attachments/assets/ca4a0aa7-bad0-4c17-a9f9-ad3fa9627448

PR #1441, opened 2025-09-12
Attributes clipboard conversion between different types of fields

Maksims:
Fixes partially #1369

Allows to copy/paste values between fields of different types, where it makes sense.
Also allows to copy assets from the assets panel, and paste in attributes.

## Some highlights:
1. Strings can be parsed to rgba/rgba and vice-versa.
2. Vectors of different dimensions can be copied to/from.
3. Asset field of specific asset type can be copied around assets without specific type.
4. Asset field of non specific type can be copied to asset field with specific type if type of copied asset matches.
5. Arrays of strings to tags, or string, and text, and vice-versa between them.

https://github.com/user-attachments/assets/4d5930df-f9f6-45bf-b990-e532afead666

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,7 +42,7 @@
         "monaco-themes": "^0.4.2",
         "nodemon": "3.1.7",
         "ot-text": "github:playcanvas/ot-text",
-        "playcanvas": "^2.10.6",
+        "playcanvas": "^2.11.6",
         "postcss": "8.4.48",
         "rollup": "4.25.0",
         "rollup-plugin-copy": "3.5.0",
@@ -6110,9 +6110,9 @@
       }
     },
     "node_modules/playcanvas": {
-      "version": "2.10.6",
-      "resolved": "https://registry.npmjs.org/playcanvas/-/playcanvas-2.10.6.tgz",
-      "integrity": "sha512-u/mHmG0Vk6Nr7qmMNCiAs1N2uQM/xjRHZ2fDcGJTpSCDU1dN3p1Nb3eGriQmJmJbpH1Y7D560JSWRnqavqUm7w==",
+      "version": "2.11.6",
+      "resolved": "https://registry.npmjs.org/playcanvas/-/playcanvas-2.11.6.tgz",
+      "integrity": "sha512-hyr5JiU1JSefUT2/HxZz5KPc8FPsu935hpfnw08Xe1H26P36tezzcbvq6qw3DuIOaoBXcpoJqOr9XrKVYQB0ag==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "monaco-themes": "^0.4.2",
     "nodemon": "3.1.7",
     "ot-text": "github:playcanvas/ot-text",
-    "playcanvas": "^2.10.6",
+    "playcanvas": "^2.11.6",
     "postcss": "8.4.48",
     "rollup": "4.25.0",
     "rollup-plugin-copy": "3.5.0",
--- a/src/editor/inspector/components/script.ts
+++ b/src/editor/inspector/components/script.ts
@@ -1,3 +1,4 @@
+import type { AssetObserver } from '@playcanvas/editor-api';
 import { Panel, Container, Button, BooleanInput, LabelGroup, Label, SelectInput, BindingTwoWay, ArrayInput } from '@playcanvas/pcui';
 
 import { ComponentInspector } from './component.ts';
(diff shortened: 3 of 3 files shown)


src/editor/storage/clipboard-context-menu.ts
```diff
@@ -37,6 +36,531 @@ for (const type of assetTypes) {
...
+        }
+    ], [
+        'rgb-string',
+        (n, o) => {
+            return pc.Color.prototype.toString.call({
+                r: n[0],
+                g: n[1],
+                b: n[2],
+                a: 0
+            }, false).toUpperCase();
```
kpal81xd: Using prototype is a bit hacky. Prob better to use a temporary color with a utility function
```ts
const color = new pc.Color();
const rgba2hex = (r, g, b, a) => color.set(r, g, b, a).toString(false).toUpperCase();
```

src/editor/assets/assets-copy.ts
```diff
@@ -11,6 +11,16 @@ editor.once('load', () => {
             assets: assets.map(asset => asset.get('id'))
         };
 
+        const assetClipboardIds = assets.filter((asset) => {
+            return asset.get('type') !== 'folder';
+        }).map((asset) => {
+            return asset.get('id');
+        });
```
kpal81xd:
```suggestion
        const assetClipboardIds = assets.reduce((ids, asset) => {
            if (asset.get('type') !== 'folder') {
                return ids;
            }
            ids.push(asset.get('id'));
            return ids;
        }, []);
```

src/editor/storage/clipboard-context-menu.ts
```diff
@@ -37,6 +36,619 @@ for (const type of assetTypes) {
...
+// converts string to vector (2-4 components)
+// these formats supported:
+// 8, 16, 32
+// [ 8, 16 ]
+// { "x": 8, "z": 32 }
+const convertStringToVec = (string, valueOld) => {
+    const valueNew = [];
+    for (let i = 0; i < valueOld.length; i++) {
+        valueNew[i] = valueOld[i];
+    }
```
kpal81xd:
```suggestion
    const valueNew = valueOld.slice();
```

src/editor/storage/clipboard-context-menu.ts
```diff
@@ -230,8 +834,33 @@ editor.once('load', () => {
         }
 
         // types should match
+        // or there should be a valid conversion option
         if (paste.type !== type) {
-            return false;
+            if (!convertTypes.has(`${paste.type}-${type}`)) {
+                return false;
+            } else if (paste.type === 'asset' && type.startsWith('asset:') && paste.value) {
+                const assetType = editor.call('assets:get', paste.value)?.get('type');
```
kpal81xd:
```suggestion
            if (!convertTypes.has(`${paste.type}-${type}`)) {
                return false;
            }
            if (paste.type === 'asset' && type.startsWith('asset:') && paste.value) {
                const assetType = editor.call('assets:get', paste.value)?.get('type');
```

src/editor/storage/clipboard-context-menu.ts
```diff
@@ -37,6 +36,619 @@ for (const type of assetTypes) {
...
+                bytes = pc.math.intToBytes24(i);
+                bytes[3] = 255;
+            }
+
+            valueNew[0] = bytes[0] / 255;
+            valueNew[1] = bytes[1] / 255;
+            valueNew[2] = bytes[2] / 255;
+            if (valueOld.length === 4) {
+                valueNew[3] = bytes[3] / 255;
+            }
```
kpal81xd: You can use `fromString` method on to parse hex into `pc.Color`

src/editor/storage/clipboard-context-menu.ts
```diff
@@ -37,6 +36,619 @@ for (const type of assetTypes) {
     types.add(`array:asset:${type}`);
 }
 
+// converts string to vector (2-4 components)
+// these formats supported:
+// 8, 16, 32
+// [ 8, 16 ]
+// { "x": 8, "z": 32 }
+const convertStringToVec = (string, valueOld) => {
```
kpal81xd:
```suggestion
const convertStringToVec = (string: string, valueOld: number[]) => {
```

src/editor/storage/clipboard-context-menu.ts
```diff
@@ -37,6 +36,619 @@ for (const type of assetTypes) {
...
+    return valueNew;
+};
+
+// converts string to color (with optional alpha)
+// these formats supported:
+// #ffffffff
+// 8, 16, 32
+// [ 8, 16 ]
+// { "x": 8, "z": 32 }
+const convertStringToColor = (string, valueOld) => {
```
kpal81xd:
```suggestion
const convertStringToColor = (string: string, valueOld: number[]) => {
```

src/editor/storage/clipboard-context-menu.ts
```diff
@@ -1,6 +1,9 @@
 import { Menu, MenuItem, Label } from '@playcanvas/pcui';
+import { Color } from 'playcanvas';
 
 
+const colorA = new Color();
```
kpal81xd:
```suggestion

const colorA = new pc.Color();
```

Engine isn't fully bundled in right now so use pc global

abstrakt8:
@Maksims @kpal81xd  Let's merge it in, so we can create a minor release today?  I don't have perms to update the branch and merge it in

PR #1442, opened 2025-09-15
feat(templates): implement "Unlink From Template" for multi-selection

### LINKED ISSUE #1222: "Unlink From Template" - with multi-selection does not work
In the hierarchy, sometimes it is required to unlink from templates multiple entities. But with multi-selection whole "Templates" context menu group is disabled.

It would be useful to have at least the "Unlink" feature enabled in multi-selection cases.

abstrakt8:
Fixes #1222. 

### What has changed?
* Implemented "Unlink From Template" feature for multiple entities selection
* This also required some changes in the contextual menu to enable this feature for multi-selection

### Testing

* Also tested my changes against the "Revert" in the diff overview that also uses `templates:unlink` ([here](https://github.com/playcanvas/editor/blob/5540303e6bf923fee1f4440353e3db406bcba56f/src/editor/templates/revert-overrides.ts#L296-L296))

https://github.com/user-attachments/assets/966147b6-2742-45f1-985e-f0a2d62b2252

### DIFF
--- a/src/editor/templates/unlink-template.ts
+++ b/src/editor/templates/unlink-template.ts
@@ -1,5 +1,5 @@
-// TODO: Better types once available
-type Entity = any;
+import type { EntityObserver } from '@playcanvas/editor-api';
+
 type SnapshotEntry = {
     entityId: any;
     templateId: any;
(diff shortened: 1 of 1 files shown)


src/editor/templates/unlink-template.ts
```diff
@@ -1,64 +1,89 @@
+// TODO: Better types once available
+type Entity = any;
```
kpal81xd: You can import EntityObserver type from editor API
```ts
import type { EntityObserver } from '@playcanvas/editor-api';
```

src/editor/templates/unlink-template.ts
```diff
@@ -1,64 +1,89 @@
...
+                entity.history.enabled = history;
+            });
         }
 
         function redo() {
-            entity = entity.latest();
-            if (!entity) {
-                return;
-            }
+            snapshots = [];
```
kpal81xd: I would just clear the existing reference after use instead of creating a new one. Just do `snapshots.length = 0` in undo after reapplying changes

abstrakt8: Thanks, fixed it :)

src/editor/templates/unlink-template.ts
abstrakt8: Thanks, fixed it :)

Maksims:
Awesome! It will save so much time for our workflow! Thank you! 🙏

PR #1453, opened 2025-09-17
fix(asset-panel): do not auto-select newly created asset if it would disrupt workflow

### LINKED ISSUE #1063: Do not auto-select Script Asset on creation
When creating script asset, there is a significant delay, then it auto-selects it. Often this is not desirable and feel intrusive.
There is a workflow where this is not desirable behaviour:

1. Select an Entity
2. Add Script Component
3. Right click in Assets Panel
4. Select New Asset > Script
5. Type a Script Name

**Issue:** after some delay it creates an asset and selects it.
**Expected:** see script asset to be created, preserve the current selection, so developer can type script name on inspector panel in script component block.

abstrakt8:
Fixes #1063 

### What has changed?

* Newly created assets are not focused anymore, if the user was focused on an entity just like as described in the workflow in #1063. Otherwise, if the user some other inspector panel open (e.g. asset or settings), the focus is still performed because it still gives a better UX (e.g. assurance that the file was created just like in any file explorer).

https://github.com/user-attachments/assets/a0f7af8e-972a-4161-8a3d-240114627263

### DIFF
--- a/src/editor/assets/assets-context-menu.ts
+++ b/src/editor/assets/assets-context-menu.ts
@@ -224,12 +224,21 @@ editor.once('load', () => {
                     }
                 }
 
-                const selectAsset = (asset) => {
-                    // Reason for skipping the selection can be read here: https://github.com/playcanvas/editor/issues/1063
-                    const isFocusedOnEntity = editor.api.globals.selection.items.some(item => item instanceof Entity);
-                    if (!isFocusedOnEntity) {
-                        editor.api.globals.selection.set([asset]);
-                    }
+                const selectAsset = () => {
+                    let selectionChanged = false;
+                    editor.api.globals.selection.once('change', () => {
+                        selectionChanged = true;
+                    });
+                    return (asset) => {
+                        const isFocusedOnEntity = editor.api.globals.selection.items.some(item => item instanceof Entity);
+
+                        // Reason for skipping the selection can be read here: https://github.com/playcanvas/editor/issues/1063
+                        const wouldDisruptWorkflow = isFocusedOnEntity || selectionChanged;
+
+                        if (!wouldDisruptWorkflow) {
+                            editor.api.globals.selection.set([asset]);
+                        }
+                    };
                 };
 
                 if (key === 'upload') {
@@ -242,7 +251,7 @@ editor.once('load', () => {
                             editor.call('assets:create:script', {
                                 filename: filename,
                                 parent: folder
-                            }, selectAsset);
+                            }, selectAsset());
                         });
                     }
                 } else {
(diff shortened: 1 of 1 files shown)


abstrakt8:
@Maksims I also implemented the "skipping behaviour" for other asset types (JSON, Shader, ...), should be okay right?

I'm also going to take a look at the asset creation time. From some local testing, it takes like 0.5-2.5s end-to-end for anything that is not a "Folder".

Maksims:
Thank you for working on this one, this has been forever contributor on annoyances.

Although, I can suggest to do it slightly differently: instead of checking if selection contains entities, subscribe to selection change event, and if that have happened since asset creation request was made, then prevent auto-selection. That way user can sit, wait and it will select newly created asset (eventually). Or go and do own stuff, without intrusions on selection.

When user creates from context menu in asset panel - that is actually fine to auto-select immediately.
But if we are creating from script component input field, then auto-selection - is the intrusive one.

Due to how long it takes to create, it will appear as "nothing is happening", so at least some tiny indicator that we are waiting for asset creation - will be a good help for user to know "it is going on".

abstrakt8:
> When user creates from context menu in asset panel - that is actually fine to auto-select immediately.
> But if we are creating from script component input field, then auto-selection - is the intrusive one.

Sorry, I'm a bit confused on this one: Auto-selection doesn't seem to happen if I create a script from the script component panel and in the issue (#1063 ) you were mentioning the asset panel specifically. Here is what I see in Prod and how I envision your workflow is disrupted:

https://github.com/user-attachments/assets/3faf84a0-9a36-470d-af8e-eed2e5f4c9ee

> Due to how long it takes to create, it will appear as "nothing is happening", so at least some tiny indicator that we are waiting for asset creation - will be a good help for user to know "it is going on".

Yeah good idea, though I'm going to take a look why it's taking so long in the first place and see if it's a quick fix. Otherwise, we can add a loading indicator or other UX patterns to improve the waiting behaviour

Maksims:
> Sorry, I'm a bit confused on this one: Auto-selection doesn't seem to happen if I create a script from the script component panel and in the issue (#1063 ) you were mentioning the asset panel specifically. Here is what I see in Prod and how I envision your workflow is disrupted:

Sorry, you are correct.
It only does it when you create asset from the asset panel.

But the idea to prevent selection, if selection has changed since creation request - still stands.

> Yeah good idea, though I'm going to take a look why it's taking so long in the first place and see if it's a quick fix. Otherwise, we can add a loading indicator or other UX patterns to improve the waiting behaviour

That would be an ideal outcome - quick creation, would basically solve most of the issues here.

abstrakt8:
> But the idea to prevent selection, if selection has changed since creation request - still stands.

Yeah, I like that idea and will include that in the next revision! I'll keep the check for the entity inspector being open though because user could also not be selecting anything but expect to continue with the entity inspector.

Maksims (approved):

willeastcott (approved):
Nice! 👏

PR #1467, opened 2025-09-24
When sorting assets, folders now will be at the top

### LINKED ISSUE #178: Assets list to show folders first and then files.
In Code Editor folders are shown first, and then files - this is easy to navigate and looks clean.
In Editor using list view in assets panel, does not shows folders first, it mixes them with files, leading to messy navigation and difference with Code Editor.

Code Editor (good):
image

Editor (not good):
image

Maksims:
Fixes #178

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8378,9 +8378,9 @@
       }
     },
     "node_modules/tar-fs": {
-      "version": "2.1.3",
-      "resolved": "https://registry.npmjs.org/tar-fs/-/tar-fs-2.1.3.tgz",
-      "integrity": "sha512-090nwYJDmlhwFwEW3QQl+vaNnxsO2yVsd45eTKRBzSzu+hlb1w2K9inVq5b0ngXuLVqQ4ApvsUHHnu/zQNkWAg==",
+      "version": "2.1.4",
+      "resolved": "https://registry.npmjs.org/tar-fs/-/tar-fs-2.1.4.tgz",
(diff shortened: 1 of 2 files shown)


src/editor/assets/asset-panel.ts
```diff
@@ -871,6 +889,11 @@ class AssetPanel extends Panel {
             return 1;
         }
 
+        const folderDiff = this._sortByFolder(a, b);
```
abstrakt8, line 890: Should the folder be on the top when sorting by file size? For instance, the folders would be "in the way" for users who want to look at large file sizes:

Image

Maksims: I believe it should, as it is always consistent and expected if it is same behavior for all cases.

src/editor/assets/asset-panel.ts
```diff
@@ -823,6 +823,13 @@ class AssetPanel extends Panel {
         editor.call('assets:paste', this.currentFolder, keepFolderStructure);
     }
 
+    // Sorts assets to have folders above and other files below
+    _sortByFolder(a, b) {
+        const folderA = a.get('type') === 'folder' ? 1 : 0;
+        const folderB = b.get('type') === 'folder' ? 1 : 0;
```
kpal81xd:
```suggestion
        return +(b.get('type') === 'folder') - +(a.get('type') === 'folder');
```

kpal81xd (approved):
Nice! Minor comment not critical.

PR #1469, opened 2025-09-24
Assets Panel header makeup - tooltips, styling

### LINKED ISSUE #74: Assets List - feedback
Assets List - is a long awaited feature, and it definitely will be favourite way of interacting with assets panel for many. **Great addition!**

Understanding it is not yet finalised, I would still provide some feedback and wishes, that I believe would make it even better:

1. **Design**, UI in Editor is simple, flat and functional for productivity and ease of perception. Here is concept of simple and functional design, being minimal is also clear of what is interactive on it.
image
2. When search is inserted, highlight search box clearly, to indicate it has value. Similar to search in hierarchy. Reasoning: it is easy to insert something, forget about it, and then wander why all folders are empty. Caught myself with it many times. Same goes for type filter selector. Also note X button on asset type filter, clicking specifically on which - it will clear that filter (set to All).
image
3. Scroll bar to start from below column headers, as currently there is a way to use arrows (up/down) and it would not scroll to most top item, as it is visible, but right behind column headers.
4. Arrow controls: up/down - when using it, works great if only files are in view, but once it hovers over folder, looks like next arrow keypress is overriden by left files tree, which will select next folder there.
5. Arrow left - to go "up folder" same as backspace key.
6. Arrow right - when hovering over folder, to navigate into that folder.
7. Do not stop arrow keys folder navigation when going between folders.
8. Render all folders, always at the top of the list, regardless of filters, search and sort by column.
9. When searching using RegExp, like: `*\.(jpg|png)`, respect folder selection as well, like searching with normal text.
10. **Big one:** currently, when selecting asset pickers in material or any other place, it will automatically navigate in assets panel to that asset. This is often not desirable, and feels like "magical" functionality. Would be best to simply add a button, on top right of inspector panel, close to "previous asset" button, by clicking which, would navigate in assets panel to. Please check other Editors, that have similar functionality.
11. "unused" mark, seems to be gone, is it coming back in another form? I used it a lot, and think many people did too, to figure out what is actually used in scene and not. Very useful when changing assets to non-preload, without need to check for "References".
12. Tooltips for assets panel buttons, just like on left menu, would be useful for new users.
13. Font on asset type filter to be not bold, just `font-weight: normal`

### Few questions:
1. Is it planned split view (2 lists?).
2. Custom columns, like "preload", or settings to allow setting it?

Maksims:
Partially Fixes #74 

Before:  
image

After:  
image

Tooltips when hovering:

Before:  
image

After:  
image

### This PR addresses a few things with Assets Panel header:

1. Styling for buttons to make them more like a left toolbar - simple, minimal, straight to the point.
2. Select field and Search field - are now also more minimal, and don't use "bold" font uniquely, and looks closer like other searches across the Editor.
3. Type filter as well as search bar will stay dark and apparent when not empty, so you won't be wandering "where are all my assets?" only to realize you've left something in these filters.
4. Tooltips, make them quick, fast, and clear.
5. Tooltips now respect if a button is disabled, like on the left toolbar, and have label dimmed.

@kpal81xd I've re-used one legacy tooltip for a whole header based on your recent PR, so it does not create thousands of them. 🙏

### DIFF
--- a/sass/editor/_editor-assets-panel.scss
+++ b/sass/editor/_editor-assets-panel.scss
@@ -660,25 +660,23 @@
             flex-direction: row;
             flex-grow: 1;
             align-items: center;
-        }
-
-        .pcui-button {
-            height: 24px;
-            line-height: 24px;
-            border: 1px solid $bcg-darkest;
-            background-color: $bcg-dark;
-            margin: $element-margin;
-            flex-shrink: 0;
+            height: 32px;
+            border-left: 1px solid $bcg-darkest;
         }
 
         .pcui-asset-panel-btn-small {
             padding: 0;
-            width: 24px;
+            margin: 0;
+            border: 0;
+            width: 32px;
+            height: 32px;
+            line-height: 32px;
             font-size: 15px;
-            line-height: 15px;
-            margin-left: 0;
-            margin-top: 4px;
-            margin-bottom: 4px;
+
+            &:hover {
+                z-index: 1;
+                background-color: #20292b;
+            }
         }
 
         .pcui-asset-panel-btn-active {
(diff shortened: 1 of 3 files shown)


kpal81xd (approved):
Nice! I want to get all tooltips consistent if possible makes things a lot easier to manage

Maksims:
> Nice! I want to get all tooltips consistent if possible makes things a lot easier to manage

Updating behavior (no delay) of the new tooltip and its styling to match previous one, probably should be done as a separate PR.
Then tooltips (old ones) could be updated.

Issue #1465, opened 2025-09-23
Previewing Templates


abstrakt8:
It'd be great to have a preview of the template on the side of the template inspector. The editor does it similarly for `Render` in the [render-preview.ts](https://github.com/playcanvas/editor/blob/main/src/editor/inspector/assets/render-preview.ts).

<img width="1111" height="847" alt="Image" src="https://github.com/user-attachments/assets/fbfcf4c4-ccbe-4db6-9e3c-8c8c0c666155" />

The way almost all "previewers" function is to use a corresponding thumbnail renderer that would load the assets, setup a scene and render it into a canvas. Therefore, we should also be able to get template thumbnails as a byproduct (#783). However, we need to check if performance would suffer because some templates could be very complex and we are generating the thumbnails on-the-fly on the client side.  Hence, I would like start off with a template preview in the template inspector.

Maksims:
That would definitely be useful, but as you've mentioned - generating these thumbnails should be done with care. In Assets Panel - thumbnails could be very useful, like with materials and render assets also.
But for generating them, I would ensure that only a certain amount of time a frame could be spent on generating those, and if timecap is reached, postpone the other thumbnails for the next frame and so on. So we have "async" generation, and it will not freeze, like it does now with materials (shaders) for example.

Also, there are templates that don't have anything to render, like UI or some script based stuff, so it would be good to detect if there is nothing rendered - in such case, show an icon as is now.

OlegGedzjuns:
This is amazing

Issue #783, opened 2022-06-04
Template Icon


KingMaks:
Hi, Guys

Would be cool to have preview icons on templates in assets section, so we could know which models exactly combined in any of the templates, if there is a lot of templates you can see difference only in their names because there is default icon for any template. A bit more info I explained here:
https://forum.playcanvas.com/t/template-icon/26028

Thanks!

yaustar:
We did consider generating the preview for templates but due to the potential complexity, it could be very expensive to generate unless we cache thumbnails which we don't currently do.

yaustar:
I do like the idea from the forum about being able to choose your icon. We now have the ability to exclude assets from the project build/launch so this could work 🤔

KingMaks:
> I do like the idea from the forum about being able to choose your icon. We now have the ability to exclude assets from the project build/launch so this could work 🤔

Yeas, that would be cool, since there is no necessary to have an icon for any template but at least for some of them, so choose own icons is good decision for the begining. To tell the truth that was the first thing which I tried when I stuck in that issue, I started to find how can I change default icon for template :)

willeastcott:
@abstrakt8 After the cool work you did enabling preview rendering of Template Assets, maybe thumbnail generation is something we can support too.

abstrakt8:
Technically it's a one line change:

https://github.com/playcanvas/editor/blob/3b58f2e20fa96ce33fe6617af0088cff5c25a628/src/common/pcui/element/element-asset-thumbnail.ts#L30-L31

I think the concern was that the user can experience serious freezes if we render all the template thumbnails for larger projects. 

CC: @Maksims you offered to test it (in #1465), could you report back if you have time? Just toggling this one line from false to true should suffice. I haven't done any optimisations around it though

Maksims:
> I think the concern was that the user can experience serious freezes if we render all the template thumbnails for larger projects.

If done in distributed manner between frames, then should be reasonable. I believe the worst will be when shaders have to compile the first time for first thumbnail rendering. But correct me if I'm wrong: these thumbnails are saved statically, so it only has to be done pretty much once when template is created, and only if related materials change later?

> CC: [@Maksims](https://github.com/Maksims) you offered to test it (in [#1465](https://github.com/playcanvas/editor/issues/1465)), could you report back if you have time? Just toggling this one line from false to true should suffice. I haven't done any optimisations around it though

In general, it actually works pretty well, haven't noticed any serious delays when selecting thumbnails.
[...]

PR #1470, opened 2025-09-25
feat(inspector): Add preview for templates in the inspector

abstrakt8:
Implements #1465 

## TODO
 - [ ] Watch for template changes
 - [ ] Watch for material changes

### Not in scope
* Template thumbnails (#783) due to uncertanity of how it would run. However, you can help me test by setting this line to true: https://github.com/playcanvas/editor/blob/793d204f3039e8a969f2dc03acaeddda426e3613/src/common/pcui/element/element-asset-thumbnail.ts#L29

### DIFF
--- a/src/common/pcui/element/element-asset-thumbnail.ts
+++ b/src/common/pcui/element/element-asset-thumbnail.ts
@@ -8,6 +8,7 @@ import { MaterialThumbnailRenderer } from '../../thumbnail-renderers/material-th
 import { ModelThumbnailRenderer } from '../../thumbnail-renderers/model-thumbnail-renderer.ts';
 import { RenderThumbnailRenderer } from '../../thumbnail-renderers/render-thumbnail-renderer.ts';
 import { SpriteThumbnailRenderer } from '../../thumbnail-renderers/sprite-thumbnail-renderer.ts';
+import { TemplateThumbnailRenderer } from '../../thumbnail-renderers/template-thumbnail-renderer.ts';
 import { buildQueryUrl } from '../../utils.ts';
 import { CLASS_MULTIPLE_VALUES } from '../constants.ts';
 
@@ -24,7 +25,8 @@ const CANVAS_TYPES = {
     'material': true,
     'model': true,
     'sprite': true,
-    'render': true
+    'render': true,
+    'template': false // TODO: Change to true after testing with optimisations (GH Issue #783)
 };
 
 type AssetThumbnailArgs = {
@@ -240,6 +242,9 @@ class AssetThumbnail extends Element {
                 break;
             case 'render':
                 this._canvasRenderer = new RenderThumbnailRenderer(asset, this._domCanvas);
+                break;
+            case 'template':
+                this._canvasRenderer = new TemplateThumbnailRenderer(asset, this._domCanvas);
                 break;
         }
 
--- a/src/common/thumbnail-renderers/template-thumbnail-renderer.ts
+++ b/src/common/thumbnail-renderers/template-thumbnail-renderer.ts
@@ -1,0 +1,346 @@
+import { Asset as EditorAsset } from '@playcanvas/editor-api';
+import type { Entity, Layer, Template, Asset, AppBase, WebglGraphicsDevice, MeshInstance, RenderComponent, BoundingBox, AssetReference } from 'playcanvas';
+
+import { ThumbnailRenderer } from './thumbnail-renderer.ts';
+
+function calculateBoundingBoxOfMeshInstances(meshInstances: MeshInstance[]): BoundingBox {
+    const aabb = new pc.BoundingBox();
+    let first = true;
+    for (const meshInstance of meshInstances) {
+        if (first) {
+            aabb.copy(meshInstance.aabb);
+            first = false;
+        } else {
+            aabb.add(meshInstance.aabb);
+        }
+    }
+    // Nothing there, fall back to a default cube
+    if (first) {
+        aabb.center.set(0, 0, 0);
+        aabb.halfExtents.set(0.1, 0.1, 0.1);
+    }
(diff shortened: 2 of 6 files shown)


Maksims:
One significant think to look at, is that shader compilation is inevitable in this case. And it can be of serious problem in projects with more complex setups and combinations of materials.
We already have a problem where navigating to a folder with 10 materials, freezes browser for 5+ seconds.

To avoid that, we would need to ensure not too many thumbnails are ever attempted to be rendered in a single frame.

src/common/thumbnail-renderers/template-thumbnail-renderer.ts
```diff
@@ -0,0 +1,422 @@
...
+    private materialWatches: Map<number, number> = new Map();
+
+    private readonly handleQueueRender: () => void;
+
+    private readonly handleTemplateAssetLoad: (asset: Asset) => void;
+
+    private readonly handleTemplateAssetChange: (asset: Asset) => void;
+
+    constructor(editorAsset: EditorAsset,
+                private _canvas: HTMLCanvasElement) {
```
kpal81xd: Can you not just omit the the second parameter or does it give a TS error

src/common/thumbnail-renderers/thumbnail-renderer.ts
```diff
@@ -1,10 +1,9 @@
-const pc = typeof window.pc !== 'undefined' ? window.pc : {
-    Layer: class {},
-    RenderTarget: class {},
-    LayerComposition: class {}
-};
```
kpal81xd, line 5: Thumbnail renderer gets loaded when loading PCUI in Editor blank page so need to keep this

src/common/thumbnail-renderers/template-thumbnail-renderer.ts
```diff
@@ -0,0 +1,422 @@
...
+        this.scene.cameraEntity.camera.renderTarget = renderTarget;
+        this.scene.cameraOrigin.setLocalEulerAngles(rotationX, rotationY, 0);
+
+        // Set Camera Position based on AABB of the mesh instances
+        const boundingRadius = aabb.halfExtents.length();
+
+        // Default (vertical) FOV = 45 Degrees.
+        // Radius/Distance = tan(FOV/2)
+        // Distance = Radius/tan(22.5) = Radius * 2.414
+        const cameraDistance = boundingRadius * 2.414 * 1.05;
```
kpal81xd: Might as well just use a constant at the top of the file `const INV_TAN_22_5 = 1 / Math.tan(Math.PI / 8)`

src/editor/inspector/assets/template-preview.ts
```diff
@@ -0,0 +1,138 @@
...
+            this._previewRenderer = undefined;
+        }
+
+        if (this._renderFrame) {
+            cancelAnimationFrame(this._renderFrame);
+            this._renderFrame = null;
+        }
+    }
+
+    _toggleSize() {
```
kpal81xd: Make these private/protected?

src/common/thumbnail-renderers/thumbnail-renderer.ts
abstrakt8: Got it, will keep it and add a comment!

src/common/thumbnail-renderers/template-thumbnail-renderer.ts
abstrakt8: Yeah, I moved it to a separate function `calculateCameraDistance` with these comments for those who are interested how this magic number was derived. This magic number `2.5` was copy pasted multiple times across all the different thumbnail renderers

src/common/thumbnail-renderers/template-thumbnail-renderer.ts
abstrakt8: The `private` is a shorthand that ensures that it's a member of the class. We also use it somewhere else

abstrakt8: I'm gonna rename it to `canvas` to reduce confusion

willeastcott:
@abstrakt8 With Engine 2.12.x released, I guess this can be merged now?

src/common/pcui/element/element-asset-thumbnail.ts
```diff
@@ -24,7 +25,8 @@ const CANVAS_TYPES = {
     'material': true,
     'model': true,
     'sprite': true,
-    'render': true
+    'render': true,
+    'template': false // TODO: Change to true after testing with optimisations (GH Issue #783)
```
kpal81xd, line 29: Has this been tested?

src/common/thumbnail-renderers/template-thumbnail-renderer.ts
```diff
@@ -0,0 +1,428 @@
...
+
+    templateInstance?: Entity;
+
+    cameraEntity: Entity;
+
+    cameraOrigin: Entity;
+
+    lightEntity: Entity;
+
+    private renderComponents: RenderComponent[] = [];
```
kpal81xd: Use underscores for private variables

abstrakt8:
Still looks good after merging in from main

src/common/pcui/element/element-asset-thumbnail.ts
abstrakt8: Thumbnails not in scope of this PR

abstrakt8: Cause I think it causes issues on larger projects

src/common/thumbnail-renderers/template-thumbnail-renderer.ts
abstrakt8: Done!

PR #1487, opened 2025-10-09
Light falloff attribute fix

Maksims:
Currently, if light is using area light shape (non punctual), then falloff property is disabled.
But if you set light shape to non-punctual and then switch off area light for the project, then falloff property will not be visible.

This PR fixes that.


willeastcott:
Nice, thanks for the fix! 🙏

PR #1489, opened 2025-10-09
Ensure select-input highlight is snappy and responsive

Maksims:
When user interacts with UI, most Editor's elements appear as soon as possible, and allow user to visually perceive and interact with them as fast as possible. This is especially relevant to power users.
So in most cases, animations like fading, highlighting, showing etc, is instant in Editor, and we want to keep it that way.
Any animation that leads to reduced experience, should not be added.
Animations that do not reduce the responsiveness of UI - are reasonable.

This PR removes fading on select fields, which makes select field to feel "laggy" where highlight is lagging behind mouse hovering.


willeastcott:
Should this just be a PCUI change?

Maksims:
> Should this just be a PCUI change?

I don't really know as there might be other developers using pcui, and for them it might be ok to have transition.

willeastcott:
> I don't really know as there might be other developers using pcui, and for them it might be ok to have transition.

But you're saying it's objectively better without the transition. So the SelectInput here, for example, should be better with your change, right?

https://developer.playcanvas.com/user-manual/pcui/examples/

Which suggests to me that updating this would make more sense:

https://github.com/playcanvas/pcui/blob/main/src/components/SelectInput/style.scss#L135

Maksims:
> But you're saying it's objectively better without the transition. So the SelectInput here, for example, should be better with your change, right?

Absolutely. Power tools needs to be snappy and responsive. Animations that lag behind - make "laggy" appearance. Especially when they interfere with "communication" of what user have hovered.

> Which suggests to me that updating this would make more sense:
> 
> https://github.com/playcanvas/pcui/blob/main/src/components/SelectInput/style.scss#L135

I will make a PR there.

Maksims:
Subjectively, I would remove that kind of fading on everything that is hoverable, this includes any input fields, even for text. That "fading" is totally unnecessary there.

PR #1504, opened 2025-10-30
fix(inspector): use correct tooltip for thicknessVertexColorChannel

aidinabedi:
Currently, `thicknessVertexColorChannel` incorrectly uses the tooltip for `thicknessVertexColor`, instead of the channel. And the `thicknessVertexColorChannel` tooltip itself has incorrect sub-title `{Boolean}`, instead of `{String}`. This PR fixes both.

Before PR:

chrome_Rtn0eBjYxh

After PR:

chrome_nRQfYtwgKj


willeastcott:
Well spotted - thanks for the fix! 🙏

PR #1532, opened 2025-11-17
Maintenance message updates

kpal81xd:
### Changes
- Updated message to explicitly say click here for more info.

### Preview

image

### DIFF
--- a/src/editor/toolbar/toolbar-maintenance.ts
+++ b/src/editor/toolbar/toolbar-maintenance.ts
@@ -26,7 +26,7 @@ editor.once('load', async () => {
 
     const label = new Label({
         unsafe: true,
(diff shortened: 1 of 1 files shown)


Maksims:
Is it possible to remember closed state per user (not per user_project), so once closed for one instance, does not come up on other projects, unless it is a new maintenance popup?
Implementation simplicity: have a single global incremental counter for maintenance popup (new popup would increment), and single user field that would write latest maintenance popup number when closed.

kpal81xd:
> Is it possible to remember closed state per user (not per user_project), so once closed for one instance, does not come up on other projects, unless it is a new maintenance popup? Implementation simplicity: have a single global incremental counter for maintenance popup (new popup would increment), and single user field that would write latest maintenance popup number when closed.

Yea should be able to just store the state in local storage

PR #1539, opened 2025-11-19
Add vertexColorGamma attribute to StandardMaterial

LeXXik:
Depends on https://github.com/playcanvas/engine/pull/8122
Also needs a schema update on the backend.

Adds `vertexColorGamma` attribute to `Other` section of StandardMaterial.

Screenshots: left cubes FBX with vertex colors, right same but in GLB format.

FBX material Vertex Color Gamma unchecked (default):

image

FBX material Vertex Color Gamma checked:

image

Note: `Vertex Color Gamma` checkbox stays unchecked on the screenshot, since the backend fails to validate the path. The material does get updated, though.


LeXXik:
@mvaligursky perhaps we could enable the property on the imported FBX material by the frontend? Not sure where to do that. Maybe @kpal81xd has an idea.

mvaligursky:
> @mvaligursky perhaps we could enable the property on the imported FBX material by the frontend? Not sure where to do that. Maybe @kpal81xd has an idea.

Yep good call, the less breaking this change is the better.

willeastcott:
Assigning this to @kpal81xd but not sure when he'll be free to help get this over the finish line. But thanks for getting this feature started @LeXXik! 🙏

mvaligursky:
ah!
That's the one I mentioned a few weeks ago that should be done before we
ship the engine.
Hmm, let's discuss tomorrow.

PR #1551, opened 2025-11-21
Absolute import support

kpal81xd:
### What's changed
- Adds absolute path support (will be useful when merging editor-api)
- Removed .ts extensions (side effect)
- Updated eslint config (side effect)

### DIFF
--- a/src/editor/inspector/assets/animation-preview.ts
+++ b/src/editor/inspector/assets/animation-preview.ts
@@ -2,7 +2,7 @@ import { Panel } from '@playcanvas/pcui';
 
 import { AnimViewer } from './anim-viewer';
 import { AssetInspectorPreviewBase } from './asset-preview-base';
-import type { Attribute } from '../attribute.type.d';
+import type { Attribute } from '../attribute.type';
 import { AttributesInspector } from '../attributes-inspector';
 
 


src/editor/inspector/assets/animation-preview.ts
```diff
@@ -1,9 +1,9 @@
 import { Panel } from '@playcanvas/pcui';
 
-import { AnimViewer } from './anim-viewer.ts';
-import { AssetInspectorPreviewBase } from './asset-preview-base.ts';
-import type { Attribute } from '../attribute.type.d.ts';
-import { AttributesInspector } from '../attributes-inspector.ts';
+import { AnimViewer } from './anim-viewer';
+import { AssetInspectorPreviewBase } from './asset-preview-base';
+import type { Attribute } from '../attribute.type.d';
```
abstrakt8: I think the file extension is `.d.ts` and we should just use:

```ts
import type { Attribute } from '../attribute.type";
```

abstrakt8: I'm not sure, but is it even necessary to have that file as `.d.ts` ? We can probably just move it to `.ts`

abstrakt8:
Reviewed all files, looks good!

PR #1563, opened 2025-11-25
Applies shader pass to all cameras

### LINKED ISSUE #1557: Render setting doesn't carry over when changing camera view
If you change the render setting and then change the camera view it doesn't keep that same render setting.

The options show it as the new setting but it's incorrect.

https://github.com/user-attachments/assets/1d270abc-8ac9-4e2b-99f9-57869603dbd5

kpal81xd:
Fixes #1557


kpal81xd:
@mvaligursky should I apply it to the scene cameras too? What do you think?

PR #1568, opened 2025-12-01
Expose FBX import option

slimbuck:
Expose an ASSET IMPORT option to resolve unique vertices using indices.

### DIFF
--- a/src/editor/assets/assets-context-menu.ts
+++ b/src/editor/assets/assets-context-menu.ts
@@ -437,23 +437,25 @@ editor.once('load', () => {
     });
 
     // replace
-    const replaceAvailable = {
-        'material': true,
-        'texture': true,
-        'textureatlas': true,
-        'model': true,
-        'animation': true,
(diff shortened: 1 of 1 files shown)


src/editor/attributes/reference/settings.ts
```diff
@@ -364,6 +364,10 @@ editor.once('load', () => {
         name: 'settings:asset-import:importMorphNormals',
         title: 'Import Morph Target Normals',
         description: 'Imports morph target normals when importing a model. Disable this if morph target normals look incorrect.'
+    }, {
+        name: 'settings:asset-import:useUniqueIndices',
+        title: 'Use Unique Indices',
+        description: 'Use vertex indices to resolve uniqueness when importing FBX models.'
```
willeastcott: I think this needs to better explain why you'd enable/disable this.

slimbuck: Any better?

PR #1460, opened 2025-09-18
Use KeyboardEvent#code in camera-fly.ts

abstrakt8:
Fixes #1451 specifically

### What has changed?
- Using `KeyboardEvent#code` ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) ) for orbit control, so that the physical keys 
- Good tool to check the different keyboard codes: https://www.toptal.com/developers/keycode

---

Will raise another PR to fix the #48 , but the current ~63 editor hotkeys don't seem to clash with the WASDQE (physical keys), so I think it's safe to merge separately

- [x] I confirm I have read the [contributing guidelines](https://github.com/playcanvas/editor/blob/main/.github/CONTRIBUTING.md)

PR #1584, opened 2025-12-03
[FIX] Fix camera fly controls for non-QWERTY keyboard layouts

### LINKED ISSUE #1451: Azerty layout issue: ZQSD vs WASD
Hello there, in France we are using Azerty keyboards so we cannot use WASD to freely move, the proper layout would have been ZQSD for us (A and E to move up/down). 

I know that we could use the arrows keys but coming from Unity there are a lot of habits (and FPS games too!). I noticed that the Playcanvas demos are having the same issues and I have to use W to move which is at the bottom of the keyboard for us.

Would it be possible to look into it ? I guess the issue is there too for other non Qwert* layouts.

Thanks again for your time

Ludovic

willeastcott:
Fixes #1451

**Problem**

Camera fly controls (WASD) didn't work properly on AZERTY keyboards because the code was detecting character values instead of physical key positions. On AZERTY, pressing the physical W key types "Z", making navigation unintuitive.

**Solution**

* Changed camera fly controls to use `KeyboardEvent.code` (physical position) instead of `KeyboardEvent.key` (character) for WASD keys
* Arrow keys continue to use character detection since they're consistent across all layouts

> [!IMPORTANT]
> Remapped "Previous Selection" shortcut from Shift+Z to Shift+B to avoid conflicts on AZERTY keyboards

**Result**

Camera movement now works consistently across all keyboard layouts (QWERTY, AZERTY, QWERTZ, etc.) following gaming industry standards where WASD position is universal.

**Testing**

Tested on QWERTY and AZERTY layouts:

✅ WASD keys work in physical position on all layouts
✅ Arrow keys continue to work
✅ No conflicts with editor shortcuts
✅ Shift+B now triggers "Previous Selection"

### DIFF
--- a/src/editor/viewport/camera/camera-fly.ts
+++ b/src/editor/viewport/camera/camera-fly.ts
@@ -22,32 +22,28 @@ editor.once('viewport:load', (app) => {
     };
 
     const keyMappings = new Map([
-        // Arrow keys (use evt.key - consistent across layouts)
-        ['arrowup', 'forward'],
-        ['arrowleft', 'left'],
-        ['arrowdown', 'back'],
-        ['arrowright', 'right'],
-        ['pageup', 'up'],
-        ['pagedown', 'down'],
-        // WASD keys (use evt.code - physical position)
-        ['keyw', 'forward'],
-        ['keya', 'left'],
-        ['keys', 'back'],
-        ['keyd', 'right'],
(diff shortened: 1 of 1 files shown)


willeastcott:
Closed in preference to https://github.com/playcanvas/editor/pull/1460

PR #1586, opened 2025-12-03
[FIX] Prevent collision component visualization disappearing when Position Offset is edited

### LINKED ISSUE #1294: Offsetting a Rigid body collider sometimes is hidden in the interface.
Some details: https://forum.playcanvas.com/t/bug-in-playcanvas-2-when-offsetting-a-rigid-body-collider/39586/6

willeastcott:
Fixes #1294 

https://github.com/user-attachments/assets/79660f49-a745-4674-b726-11436dd1cc3c

**Problem:**

When editing the Position Offset (`linearOffset`) property on a newly created collision component, the collision visualization box would disappear. Editing Rotation Offset first would make it work, or reloading the editor would fix it.

**Root Cause:**

The angularOffset property is stored in the editor as a 3-element euler angles array [0, 0, 0] but the PlayCanvas engine requires it to be a Quat object at runtime. When the component is created, if angularOffset isn't converted to a Quat, subsequent property updates fail. Specifically, when linearOffset is set, the engine's onSetOffset handler calls angularOffset.equals(Quat.IDENTITY), which fails if angularOffset is still an array instead of a Quat object.

**Solution:**

Added conversion for collision component's angularOffset property in the components:convertValue method in schema-components.ts. When an euler angles array is provided, it's converted to a Quat before being passed to the engine. This centralized conversion ensures all code paths are covered:

* Component creation (via runtimeComponentData)
* Property updates (via generic setter)
* Any other conversion path

The linearOffset property already works correctly through the existing vec3 conversion handler.

### DIFF
--- a/src/editor/viewport/camera/camera.ts
+++ b/src/editor/viewport/camera/camera.ts
@@ -136,6 +136,14 @@ editor.once('load', () => {
             }
 
             editor.emit('camera:change', currentCamera, old);
+
+            // Update shadow maps for lights with Shadow Update Mode set to "Once"
+            editor.call('entities:list').forEach((e) => {
+                if (e.get('components.light.shadowUpdateMode') === pc.SHADOWUPDATE_THISFRAME && e.entity?.light) {
+                    e.entity.light.shadowUpdateMode = pc.SHADOWUPDATE_THISFRAME;
+                }
+            });
+
             editor.call('viewport:render');
         });
 


willeastcott:
@Maksims - I'm not convinced I have the most elegant solution here. It seems to fix the problem nicely, and it eliminates some other 'special casing' code, but I was wondering if you had an opinion.

Maksims:
Hm, not sure we are storing quats anywhere else apart of this place. Entity has quat for rotation, but it is stored as euler and is set as euler, should this follow the same logic?

willeastcott:
@Maksims This is how rotation is set:

https://github.com/playcanvas/editor/blob/main/src/editor/viewport/viewport-entities-create.ts#L35

And:

https://github.com/playcanvas/editor/blob/main/src/editor/viewport/viewport-entities-observer-binding.ts#L22

I wasn't able to get this working in the 'binding' files....but handling it there means having to address multiple code paths and duplicating the 'special casing'. And there is already a precedent in the `convertValue` function handling `batchGroupId`.

willeastcott:
TBH, this fix works and is incredibly low risk so I'm leaning towards just merging it...and if we want to refactor later, that's totally fine. But if you have a specific change you'd like me to make, I'm open to suggestion.

Maksims:
I guess this is fine, as it is very isolated case.

PR #1593, opened 2025-12-06
[FIX] Fix nested entity enabled/disabled visual state in hierarchy

### LINKED ISSUE #1508: Nested entities enabled/disabled state can be stuck or incorrect in hierarchy
1. Create entity with nested entity
2. Disable nested entity, disable parent entity
3. Enable nested entity, opacity in hiearachy will change, but should not, entity should stay visually disabled
4. Enable/disable nested entity multiple times, now entity is visually stuck in enabled state in hierarchy and doesn't react to parent state

https://github.com/user-attachments/assets/59881b29-b3b4-44ff-a864-cfe209bcf5a5

willeastcott:
Fixes the entity enabled/disabled visual state (opacity) in the hierarchy treeview to correctly reflect whether an entity is effectively enabled - considering both its own `enabled` property and all ancestors.

https://github.com/user-attachments/assets/e587d4a1-11f7-4ed0-9378-740eebb33958

**Changes:**
- Added `_isParentDisabled()` to check if any ancestor is disabled
- Added `_updateTreeItemEnabledState()` to recursively update visual state
- When an entity's `enabled` changes, all descendants now update their visual state
- When reparenting, entities update their visual state based on new parent hierarchy

**Behavior:**
- Child entities under a disabled parent now correctly appear disabled
- Toggling a child's `enabled` while parent is disabled no longer incorrectly changes its visual appearance

Fixes #1508
Fixes #1593

### DIFF
--- a/modules/editor-api/src/entities/delete.ts
+++ b/modules/editor-api/src/entities/delete.ts
@@ -159,7 +159,9 @@ async function deleteEntities(entities: Entity[] | Entity, options: { history?: 
                 });
 
                 entities.forEach((entity) => {
-                    updateReferences(entityReferences, entity.get('resource_id'), entity.get('resource_id'));
+                    entity.depthFirst((e: Entity) => {
+                        updateReferences(entityReferences, e.get('resource_id'), e.get('resource_id'));
+                    });
                 });
 
                 if (api.selection) {
--- a/src/editor/assets/assets-context-menu.ts
+++ b/src/editor/assets/assets-context-menu.ts
@@ -65,7 +65,6 @@ editor.once('load', () => {
     const notDownloadable = new Set([
         'folder',
         'sprite',
-        'animstategraph',
         'render',
         'template'
     ]);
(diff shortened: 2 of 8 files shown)


src/editor/entities/entities-treeview.ts
```diff
@@ -600,18 +600,79 @@ class EntitiesTreeView extends TreeView {
...
+            // - Our parent was disabled, OR
+            // - We ourselves are disabled
+            const childParentDisabled = parentDisabled || !selfEnabled;
+            entity.get('children').forEach((childId) => {
+                const child = this._entities.get(childId);
+                if (child) {
+                    this._updateTreeItemEnabledState(child, childParentDisabled, true);
+                }
+            });
+        }
```
slimbuck: This code could be a bit simpler and less obfuscated.

Parent state doesn't need checking if self is disabled and `childParentDisabled` is just `!this.enabled`.

```suggestion
        // Visual state is enabled only if both self is enabled AND no parent is disabled
        item.enabled = entity.get('enabled') && !(parentDisabled ?? this._isParentDisabled(entity));

        if (recurse) {
            const childParentDisabled = !this.enabled;
            entity.get('children').forEach((childId) => {
                const child = this._entities.get(childId);
                if (child) {
                    this._updateTreeItemEnabledState(child, childParentDisabled, true);
                }
            });
        }
```

PR #1598, opened 2025-12-07
Add Download GLB option for Entities in Hierarchy Panel

### LINKED ISSUE #1332: Template - download GLB
It would be amazing, if there was a capability to download a template as GLB.
GLB would contain hierarchy, related meshes, materials and textures.

A few places where it would be available:

1. Context menu on the template asset, would have an option "Download as GLB".  
2. Inspector, would have a button like on Container asset: "Download as GLB".

This feature could even be available in hierarchy, and not templates, but that would be a bit more complex to implement.

willeastcott:
- Adds a new "Download GLB" context menu option for entities in the hierarchy panel
- Uses the engine's `GltfExporter` to export the selected entity and all its children (meshes, materials, textures) to a GLB file
- Only enabled for single entity selection
- Reorders context menu to place Delete last (standard UX convention for destructive actions)

https://github.com/user-attachments/assets/d518fac2-007e-42e2-bd2e-80731f4281d3

Closes #1332

### DIFF
--- a/modules/editor-api/src/entities/delete.ts
+++ b/modules/editor-api/src/entities/delete.ts
@@ -159,7 +159,9 @@ async function deleteEntities(entities: Entity[] | Entity, options: { history?: 
                 });
 
                 entities.forEach((entity) => {
-                    updateReferences(entityReferences, entity.get('resource_id'), entity.get('resource_id'));
+                    entity.depthFirst((e: Entity) => {
+                        updateReferences(entityReferences, e.get('resource_id'), e.get('resource_id'));
+                    });
                 });
 
                 if (api.selection) {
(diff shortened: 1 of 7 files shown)


Maksims:
It is more of an "export" than "download".

src/editor/entities/entities-context-menu.ts
```diff
@@ -204,6 +204,41 @@ editor.once('load', () => {
             }
         });
 
+        menuData.push({
+            text: 'Download GLB',
```
willeastcott:
```suggestion
            text: 'Export as GLB',
```

willeastcott:
> It is more of an "export" than "download".

I've updated the menu item from `Download GLB` to `Export as GLB`.

willeastcott:
We can certainly add USDZ support. I'll leave it for a follow up PR.

OlegGedzjuns:
This is extremely usefull, thanks a lot

Issue #1376, opened 2025-08-08
Convert Legacy UI to PCUI


kpal81xd:
There still exists a lot of references to custom Legacy UI elements in the codebase. These should be migrated only use PCUI

willeastcott:
I've gradually been doing this migration over an extended period. The Code Editor is now 100% PCUI. The main Editor still has a lot of legacy UI usage. By yeah, it would be good to gradually eliminate the legacy stuff. PCUI is so awesome now, powering several great apps! ❤️

Maksims:
That tooltip is snappy, clean and precise, spawning where it should be.
There should be a decent alternative to migrate from it, as the current alternative has degraded UX and cleanness.

PR #1609, opened 2025-12-10
Migrate Builds & Publish panel to PCUI

willeastcott:
Migrates 51 calls to `Legacy` UI constructors to PCUI. Only 2 remain: `LegacyList` and `LegacyListItem` which don't have a PCUI equivalent, so will be handled in a follow up PR. Styling looks very similar to what it was previously...just more PCUI-like.

Partially addresses #1376 


kpal81xd:
Need to fix tests so will merge last

PR #1625, opened 2025-12-14
[FIX] Fix code changes lost when saving during an in-progress save

### LINKED ISSUE #813: Code changes aren't 'saved' to the launch tab if an edit and save is made during an in progress save
- Make a code change
- Save
- While saving, make another code change and save really quickly
- When launching, notice that the code used in the launch tab doesn't have the second code chang

Also note, this affects the VC checkpoints as it only checks in the first code change. This affects users more in countries far from UK because of the network latency

https://user-images.githubusercontent.com/16639049/177170098-7a9a7d61-6527-4699-9d21-7146d145a46f.mp4

- When you refresh, it shows the file as unsaved
image

willeastcott:
### Problem

When a user made code changes and saved, then quickly made additional changes and saved again while the first save was still in progress, the second save was silently blocked. This resulted in the new changes not being persisted, which was especially problematic for users with high network latency.

### Solution

Added a pending save queue that tracks save requests made during an in-progress save. When the current save completes, any pending save is automatically triggered, ensuring no changes are lost.

### Changes

- Added `pendingSave` Set to track documents that need saving after the current save completes
- Modified `editor:command:save` to queue pending saves instead of silently ignoring them
- Process pending saves on both success and error completion
- Replaced `Record<string, boolean>` with `Set<string>` for cleaner membership tracking
- Extracted `onSaveComplete` helper to reduce duplication between success/error handlers

Fixes #813

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "@playcanvas/editor-api": "file:modules/editor-api",
         "@playcanvas/eslint-config": "2.1.0",
         "@playcanvas/observer": "1.6.6",
-        "@playcanvas/pcui": "5.4.0",
+        "@playcanvas/pcui": "5.4.1",
         "@playcanvas/pcui-graph": "4.1.0",
         "@rollup/plugin-commonjs": "28.0.9",
         "@rollup/plugin-json": "6.1.0",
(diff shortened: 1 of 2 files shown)


src/code-editor/menu-panel/file/save.ts
```diff
@@ -216,39 +225,51 @@ editor.once('load', () => {
...
+
+        // If a save was requested while we were saving, trigger it now
+        if (pendingSave.has(id)) {
+            pendingSave.delete(id);
+            // Use setTimeout to allow the current save to fully complete before starting a new one
+            setTimeout(() => {
+                editor.call('editor:command:save', id);
+            }, 0);
+        }
+    }
```
kpal81xd, line 248: This only accounts for a single pending save. Better to use an asynchronous queue

willeastcott: One's all you need. You can safely discard others while one is already pending. Once the pending one is processed, it moves to a saving state where another pending save can be queued.

kpal81xd (approved):

PR #1711, opened 2026-01-29
Material migration cleanup

kpal81xd:
Closes #1710

### What's Changed
- Removes `useGamma` migration and updates binding of `useTonemap` to `useGammaTonemap`
- Updates setting of `shader` to always set to `blinn` if not already (compatibility with engine V1) 

### DIFF
--- a/src/editor/assets/assets-migrate.ts
+++ b/src/editor/assets/assets-migrate.ts
@@ -133,10 +133,9 @@ editor.once('load', () => {
         });
 
         if (asset.has('data.useGamma')) {
-            const tonemap = asset.get('data.useGamma') ?? true;
-            tonemap.write = true;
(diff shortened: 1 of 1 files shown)


src/editor/assets/assets-migrate.ts
```diff
@@ -285,8 +272,8 @@ editor.once('load', () => {
             asset.set('data.clearCoatNormalMapOffset', [0, 0]);
         }
 
-        // remove shader since physical material is now the default
-        if (asset.has('data.shader')) {
+        // NOTE: set shader to blinn if it is not already for engine v1
+        if (asset.get('data.shader') !== 'blinn') {
```
willeastcott, line 276: Why not just get once and cache?

PR #1754, opened 2026-02-13
feat: expose new engine properties in editor UI

### LINKED ISSUE #286: Rolling friction property of Rigidbody Component
If https://github.com/playcanvas/engine/pull/2785 commit goes through, the Editor would need to add a new number field property to the Rigidbody Component to allow setting a rolling friction of a rigidbody:

```
{
        name: 'rollingFriction',
        title: 'Rolling Friction',
        type: 'number', default: 0, min: 0, max: 1, precision: 4,
        description: 'Torsional friction orthogonal to contact point normal. Generally, you want this to be zero or close to zero. Defaults to 0.'
}

```

### LINKED ISSUE #986: Expose twoSidedLighting in the Material Inspector
Forum: https://forum.playcanvas.com/t/rendering-second-camera-to-texture-has-lighting-issues-for-imported-gltf-materials-when-flipping-y-on-render-target/29902

Project: https://playcanvas.com/project/1044212/overview/f-dark-render-target

Imported material from model has `twoSidedLighting` enabled and causing issues when rendering to a texture.

As far as I can see, there isn't a setting in the material inspector to disable this. 

The Editor should not be creating materials that cannot be edited in the inspector.

### LINKED ISSUE #1245: Expose PCSS (soft shadows) parameters in the Editor
New parameters for directional lights were added here, and should be exposed for engine v2
https://github.com/playcanvas/engine/pull/7258

### LINKED ISSUE #1324: Expose depthWrite option for skydome rendering
this is the engine feature: https://github.com/playcanvas/engine/pull/7729

kpal81xd:
Fixes #986, fixes #1245, fixes #1324, fixes #286

### What's Changed

Expose the following engine properties that were previously missing from the Editor UI, verified against the engine source for correct types, defaults, and property names:

**Material:**
- `twoSidedLighting` (boolean) — backface normal calculation (#986)

**Components:**
- `light.penumbraSize` (number) — PCSS soft shadow penumbra size (#1245)
- `rigidbody.rollingFriction` (number) — torsional friction for round objects (#286)

**Scene Settings:**
- `render.skyDepthWrite` (boolean) — whether the sky writes to the depth buffer (#1324)

Each property includes:
- Inspector UI attributes with correct types matching the engine
- Attribute reference definitions (tooltips/docs)
- Asset and entity migration logic for backwards compatibility
- Visibility gating via `pathExists` where engine support is conditional
- Editor-side application of `skyDepthWrite` to `scene.sky.depthWrite` (engine's `applySettings` doesn't handle it yet)

Requires a companion backend schema PR (deployed first).

----
_Created via [GitHub MCP integration](https://wiki.sc-corp.net/pages/viewpage.action?spaceKey=SEC&title=MCP+Proxy)._


mvaligursky:
- I'm not sure we want to expose detail textures at all (see comment in that PR)
- and also unified gsplat component setting

kpal81xd:
> * I'm not sure we want to expose detail textures at all (see comment in that PR)
> * and also unified gsplat component setting

Will remove those two from being exposed

src/editor/inspector/components/light.ts
```diff
@@ -589,6 +599,8 @@ class LightComponentInspector extends ComponentInspector {
             this._field(field).parent.hidden = !castShadows || shadowTypeVsm;
         });
 
+        this._field('penumbraSize').parent.hidden = !castShadows;
```
mvaligursky: penumbraSize would be only visible when the shadow type is the PCSS one.

mvaligursky (approved):
all seems ok, but added one comment that we ideally address

Maksims:
What about [light.penumbraFalloff](https://api.playcanvas.com/engine/classes/LightComponent.html#penumbrafalloff)?
And in order to configure penumbra, it only is relevant and can be configured if shadow type is PCSS: https://github.com/playcanvas/editor/blob/9b977d1e1994a6c63e5d4c3589a479149a1f77af/src/editor/inspector/components/light.ts#L265

Also, the relevant shadow settings only needs to be visible when relevant type is selected. For example falloff and penumbraSize does not need to be there if light type is not PCSS.

kpal81xd:
> What about [light.penumbraFalloff](https://api.playcanvas.com/engine/classes/LightComponent.html#penumbrafalloff)? And in order to configure penumbra, it only is relevant and can be configured if shadow type is PCSS:
> 
> https://github.com/playcanvas/editor/blob/9b977d1e1994a6c63e5d4c3589a479149a1f77af/src/editor/inspector/components/light.ts#L265
> 
> Also, the relevant shadow settings only needs to be visible when relevant type is selected. For example falloff and penumbraSize does not need to be there if light type is not PCSS.

I will add falloff in a follow up PR and recreate the release

PR #1774, opened 2026-02-20
feat: export active branch instead of always exporting main

### LINKED ISSUE #1026: Export project in Settings panel (new Editor 'Dashboard') should export the current branch
image

At the moment, the export button on this dialogue will always export main branch for archive. It would be great to have it use the current branch id instead so that users can export specific branches without REST API

willeastcott:
## Summary

image

- The Export Project button in the Project Settings panel now exports the **currently active branch** rather than always defaulting to the main branch
- When version control is enabled, the button text displays the branch name (e.g. `EXPORT PROJECT (main)`) so users know which branch will be exported
- For legacy projects without version control, behavior is unchanged (exports main branch, no branch label shown)

## Changes

- **REST API** (`src/editor-api/rest/projects.ts`): Added optional `branchId` to `ProjectExportArgs` and updated `projectExport()` to send `branch_id` in the POST body when provided
- **Editor method** (`src/editor/project/project.ts`): Updated `projects:export` to accept and forward an optional `branchId` parameter
- **UI** (`src/editor/pickers/picker-project-main.ts`): Added `getExportButtonText()` helper that appends the branch name to the button label; passes `config.self.branch.id` when exporting the current project with version control

## Test plan

- [x] Open a project with version control enabled and verify the export button shows `EXPORT PROJECT (branchName)`
- [x] Click the export button and verify the correct branch is exported (not always main)
- [x] Switch to a different branch and verify the button text updates after navigating back to Project Settings
- [x] Open a legacy project (no version control) and verify the button still shows `EXPORT PROJECT` with no branch label
- [x] Verify the loader animation works correctly (button text clears during export, restores with branch name after)

Fixed #1026

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md
--- a/sass/editor/_editor-main.scss
+++ b/sass/editor/_editor-main.scss
@@ -5584,7 +5584,7 @@ strong {
                             border-top: 1px solid $border-primary;
                         }
 
-                        .btn-view-changes {
+                        .btn-compare {
                             display: none;
                         }
                     }
--- a/src/editor/pickers/version-control/picker-version-control-checkpoints.ts
+++ b/src/editor/pickers/version-control/picker-version-control-checkpoints.ts
@@ -526,19 +526,18 @@ editor.once('load', () => {
         panelItem.flex = true;
 
         const label = new LegacyLabel({
-            text: 'Changes made since the last checkpoint'
+            text: 'Current working state of the branch'
         });
         panelItem.append(label);
 
-        // shortcut button to view changes
-        const btnViewChanges = new Button({
-            text: 'VIEW CHANGES',
+        const btnCompare = new Button({
+            text: 'COMPARE',
             size: 'small',
-            class: 'btn-view-changes'
+            class: 'btn-compare',
+            width: 110
         });
-        btnViewChanges.style.width = '110px';
-        panelItem.append(btnViewChanges);
-        btnViewChanges.on('click', () => {
+        panelItem.append(btnCompare);
+        btnCompare.on('click', () => {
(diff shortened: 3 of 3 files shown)


src/editor/project/project.ts
```diff
@@ -117,8 +117,8 @@ editor.once('load', () => {
     });
 
     // Export specified project
-    editor.method('projects:export', (projectId: string, success?: (result: unknown) => void, errorFn?: (error: unknown) => void) => {
-        editor.api.globals.rest.projects.projectExport({ projectId })
+    editor.method('projects:export', (projectId: string, branchId?: string, success?: (result: unknown) => void, errorFn?: (error: unknown) => void) => {
+        editor.api.globals.rest.projects.projectExport({ projectId, branchId })
```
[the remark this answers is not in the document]
willeastcott: The positional style matches every other method in project.ts, the method is internal with a single call-site (updated in the same commit), and an options object would be inconsistent with the existing pattern.

kpal81xd, line 121: Is the casing correct here? Referring to the comment i made in slack

willeastcott: `branchId` here is the internal TypeScript parameter name (camelCase, matching JS/TS convention). The translation to `branch_id` (snake_case, matching the REST API) happens in the REST layer in `projects.ts`:

```ts
if (args.branchId) {
    data.branch_id = args.branchId;
}
```

So the correct `branch_id` key is sent in the POST body - this is just the camelCase equivalent on the editor side.

willeastcott: See here: https://github.com/playcanvas/editor/pull/1774/changes#diff-4eb0c34ab1f60f61b07453702feab5a04fc6e46320b5087e62704ba63b2c7461R258-R260

PR #1843, opened 2026-03-03
[TEST] Enable parallel Playwright workers in CI

kpal81xd:
### What's Changed
- Set Playwright workers to 2 in CI (was hardcoded to 1), defaulting to Playwright's auto-detection locally
- Removed stale FIXME comments about account-per-worker

### Checks


kpal81xd:
Requires more testing and investigation to parallelize

PR #1864, opened 2026-03-04
fix: ensure inspector dropdowns render above console panel

### LINKED ISSUE #1275: Inspector drown down lists are rendered behind the console
Image

Example: A light component, which has the Layer combo box at the end in the inspector. Some items are not visible as those are under the console panel.

kpal81xd:
## Summary

- Inspector dropdown/select menus were rendered behind the console panel due to a z-index stacking issue
- The console (`#layout-console`) has `z-index: 1`, but the inspector panels had no z-index, causing their dropdown overlays to appear below the console
- Added `z-index: 2` to both `#layout-attributes` and `#layout-attributes-secondary` so inspector dropdowns correctly stack above the console

Fixes #1275

## Test plan

- [ ] Open the editor with the console panel visible
- [ ] Select an entity or asset that has dropdown/select fields in the inspector
- [ ] Click a dropdown near the bottom of the inspector panel where it overlaps the console
- [ ] Verify the dropdown menu renders above the console panel, not behind it
- [ ] Verify the console status bar still renders above the viewport/assets panels as intended

### DIFF
--- a/src/editor/assets/assets-panel.ts
+++ b/src/editor/assets/assets-panel.ts
@@ -49,12 +49,16 @@ editor.once('load', () => {
         assetsPanel.progressBar.value = progress * 100;
     });
 
-    // // select all hotkey
-    // // ctrl + a
+    // select all hotkey
+    // ctrl + a
     editor.call('hotkey:register', 'asset:select-all', {
         key: 'a',
         ctrl: true,
         callback: () => {
+            if (editor.call('selector:type') !== 'asset') {
+                return;
+            }
+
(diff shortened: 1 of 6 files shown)


kpal81xd:
> Okaaaay, but feels like we're masking a problem with `z-index`. I know we probably need to rethink `z-index` use across the Editor...but this is probably OK for now.

This was an automated agent based PR so still reviewing if these fixes are actually any good or not

kpal81xd:
Does not fix issue so closing in favor of improved z-index usage across codebase

Issue #1851, opened 2026-03-03
Ctrl/Cmd + A selects all entities in the scene


isumygin-sc:
Ctrl/Cmd + A selects all entities in the scene if viewport is selected - currently it works only in Assets Panel

PR #1866, opened 2026-03-04
feat: add Ctrl/Cmd+A shortcut to select all entities in scene

kpal81xd:
## Summary
- Adds a `Ctrl/Cmd+A` keyboard shortcut to select all entities in the hierarchy/viewport, addressing #1851
- Guards the existing asset `Ctrl+A` handler to only fire when the current selection type is `asset`, so the two handlers coexist without conflict
- When no selection exists (selector type is `null`), defaults to selecting all entities since the hierarchy is the primary workspace

## Test plan
- [x] Open a scene with multiple entities. Click on an entity in the hierarchy, then press `Ctrl+A` (or `Cmd+A` on Mac). Verify all entities are selected in the hierarchy panel.
- [x] Click on an asset in the assets panel, then press `Ctrl+A`. Verify all visible assets are selected (existing behavior preserved).
- [x] With nothing selected, press `Ctrl+A`. Verify all entities are selected (default behavior).
- [x] Open a picker/modal dialog, press `Ctrl+A`. Verify nothing happens (picker guard works).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/editor/entities/entities-hotkeys.ts
+++ b/src/editor/entities/entities-hotkeys.ts
@@ -148,4 +148,26 @@ editor.once('load', () => {
             }
         }
     });
+
+    // select all
+    // ctrl + a
+    editor.call('hotkey:register', 'entity:select-all', {
+        key: 'a',
+        ctrl: true,
+        callback: function () {
+            if (editor.call('picker:isOpen')) {
+                return;
+            }
+
+            const type = editor.call('selector:type');
+            if (type !== 'entity') {
+                return;
+            }
+
+            const entities = editor.call('entities:list');
+            if (entities.length) {
+                editor.call('selector:set', 'entity', entities);
+            }
+        }
+    });
 });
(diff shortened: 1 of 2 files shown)


src/editor/entities/entities-hotkeys.ts
```diff
@@ -148,4 +148,26 @@ editor.once('load', () => {
...
+    editor.call('hotkey:register', 'entity:select-all', {
+        key: 'a',
+        ctrl: true,
+        callback: function () {
+            if (editor.call('picker:isOpen')) {
+                return;
+            }
+
+            const type = editor.call('selector:type');
+            if (type === 'asset') {
```
willeastcott: Just wondering if it's safer to do `!== 'entity'` to mirror the asset one (and in case more selector types are added in the future.

willeastcott (approved):

src/editor/entities/entities-hotkeys.ts
kpal81xd: Good call — updated to `\!== 'entity'` to mirror the asset handler and be future-proof.

PR #1979, opened 2026-03-27
test: fix double-click on folders and arrow key navigation in asset picker

### LINKED ISSUE #1896: Asset picker has broken navigation
When asset picker is engaged (e.g. from script attributes), there are new regressions and issues:
1. Double click on the folder to navigate into it, now picks first asset in the folder, ending asset picker.
2. Arrows for navigating through lists, will end picker.

johnny-official:
Fixes #

### What's Changed
- TODO

### Checks

Closes #1896

### DIFF
--- a/src/editor/pickers/picker-asset.ts
+++ b/src/editor/pickers/picker-asset.ts
@@ -45,34 +45,39 @@ editor.once('load', () => {
             return;
         }
 
-        // do not allow selecting source assets
-        if (asset.get('source')) {
+        // Do not allow selecting source assets or folders for the picker result
+        if (asset.get('source') || asset.get('type') === 'folder') {
             return;
         }
 
-        if (!parseInt(asset.get('id'), 10)) {
-            // probably a legacy script
-            if (overlay.hidden ||
-                (currentType !== '*' && !Array.isArray(currentType) && currentType !== 'script')) {
-                return;
-            }
-        } else if ((currentType !== '*' && !Array.isArray(currentType) && asset.get('type') !== currentType) || asset === currentAsset) {
-            return;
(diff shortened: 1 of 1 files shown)


willeastcott:
Thanks for the contribution! I've reviewed the changes and found a few issues that would need to be addressed before this could be merged.

## Critical: `doubleclick` event does not exist

The PR adds an `assetsPanel.on('doubleclick', ...)` handler, but `AssetPanel` only emits `'select'` and `'deselect'` events. There is no `'doubleclick'` event emitter anywhere in the codebase. This means the double-click handler will never fire.

In **single-select mode**, the `select` handler now only sets `assetSelection = [asset]` without emitting `picker:asset` or closing the overlay. Confirmation is deferred to the `doubleclick` handler which never fires, so users cannot confirm a selection -- the picker is completely broken for single-select.

Consumers listening via `editor.once('picker:asset', ...)` (e.g. in `assets-context-menu.ts`) would never receive the event.

## Other issues

- **Removed `currentAsset` guard in `deselect` handler**: The original code prevented the currently-assigned asset from being deselected. Removing this changes the visual state behavior.
- **Removed deselect cleanup on hide**: The `assetSelection.forEach(asset => assetsPanel.deselect(asset))` block ensured visual selection was cleared when the picker closed. Without it, assets may remain visually selected.
- **Duplicated validation logic**: The type-validation block is copy-pasted identically in both the `select` and `doubleclick` handlers. This should be a shared helper.
- **PR metadata**: Title uses `test:` prefix but this is a fix/refactor. The body still has 'TODO' placeholders.

I'm going to close this for now. If you'd like to take another pass, the folder-filtering idea (adding `asset.get('type') === 'folder'` to the early-return guard in the `select` handler) is a reasonable starting point for fixing issue #1896.

PR #1992, opened 2026-04-09
[FIX] Rework bottom panel layout and styling

kpal81xd:
### Changes
- combine the assets and console panels into a shared tabbed bottom panel with a separate status bar
- restore the collapsed bottom-panel tab placement so it no longer overlaps the chat and profile controls
- align the bottom-panel header styling with the standard editor panels and soften the assets/console color contrast with consistent dividers

### Preview
image
image
image

### Checks


Maksims:
Please share screenshots.

kpal81xd:
> Please share screenshots.

Work in progress design - can also try it out by using the vercel deployment

Maksims:
What are the success criteria for this change?
I assume it is to reduce consumed vertical space and reduced empty wide blocks, which console takes?

After observing many developers and artists using Editor, they almost never use console. You have analytics and can see more about it. Hope you are tracking accidental "click and close straight away" there.
Console has far far less use than any of 4 main components: hierarchy, viewport, inspector and assets. So based on that, console should not occupy that much of working space.

Based on screenshot, it looks like due to change, we get reduced UX for Assets panel, which would be great to avoid.
While it makes sense that console is related to assets, with screenshot above it seems that it just moved empty space from bottom of the screen to assets panel, which is already the smallest and the most cramped panel.

Add a small button, with badges that would open the console besides "Asset Store" button. This will ensure the experience for assets panel is not reduced, and easy access to console is still provided. Additionally icon of the button can be of most important element and colored accordingly. Just icon please. We don't want always-red button in our UI.

It would open console exactly over assets panel, and have very clear way to close it.

image

kpal81xd:
@Maksims thanks for the review on the design - the idea is use the editor interface real estate more efficiently whilst also allowing to extend the panel with additional tabs in the future (inspired by Unity and VSCode layouts).

For the reduced UX in the asset panel is it because the asset action buttons are now below the panel header and the unused space to the right of the tab headers? I have tried keep the layout the same and as compact as possible

Maksims:
> For the reduced UX in the asset panel is it because the asset action buttons are now below the panel header and the unused space to the right of the tab headers? I have tried keep the layout the same and as compact as possible

Yes, exactly. Basically while hierarchy and inspector would get some extra height, assets panel would actually would not, and become even more cramoped.
Please consider the button as shown on the concept, that does not introduce extra empty panel to assets header, and does delegate appropriate and accessible place for Console button.

PR #2035, opened 2026-04-30
fix: detect and repair duplicate asset paths in auditor

### LINKED ISSUE #2030: editor.call("assets:list"); returns duplicate paths
We are listing the assets in our editor for dynamic loading with this code:

```
const allEditorAssets = editor.call("assets:list");

for (const assetRef of allEditorAssets) {
      const asset = assetRef.asset;
      if (!asset) continue;
      const pathIds = assetRef.get("path");
      ...
}
```
In some cases, pathIds contains duplicates, like: [266587536, 266587536, 286895297, 286895299]

Using this to parse a path is causing errors unless we de-duplicate it ourselves.

kpal81xd:
Fixes #2030

### What's Changed
- New `Path Audits` section in the asset auditor that flags any loaded asset whose `path` array contains duplicate folder ids and offers a one-click fix (dedupe, keeping the first occurrence).
- Companion to a pipeline-side prevention that blocks new duplicates on `asset-create` / `asset-copy` / `asset-move` but does not repair existing rows. A one-shot mongo backfill is impractical at the scale of the assets collection; this lazy, per-project repair lets each user clean up their own corrupt paths the next time they open the project.
- Small `assets-auditor` aggregator so the status badge can combine counts from multiple audit sources (sRGB + paths) cleanly. `assets-srgb` now reports through it instead of emitting `assets:auditor:issues` directly.
- Fix flows through `asset.set('path', ...)` so it goes through the standard pcui Observer → sharedb → MongoDB pipeline, and is wrapped in undo/redo.

### Preview
image

### Checks


Maksims:
Why is this even an audit, and not a proper fix with migration?
Users don't need to be involved with UI into fixing a data bug.

As you've mentioned a single one-off db query is not the way - correct.
So this should be a simple process that stream-queries the chunks of data, and processes them. Use `batchSize` with cursor instance.

PR #2036, opened 2026-04-30
feat: assign unique names to new assets within a folder

kpal81xd:
Newly created, duplicated, and pasted assets now get a unique name within their folder.

**New behavior**
- Default names are lowercase: `new.css`, `new.html`, `new.json`, `new.txt`, `new.glsl`, `folder`. Display name and filename match.
- Create / duplicate / paste in the same folder auto-suffixes: `foo (1).css`, `foo (2).css`, …
- Duplicate now works for css / html / json / text / shader (was a no-op).
- Material / sprite duplicate names go from `Foo Copy` to `Foo (1)`.
- The script-create picker now validates as you type and stays open on rejection — the inline error shows the reason (invalid filename, reserved name, or sibling collision) and the typed value is preserved so you can correct it. Previously the picker dismissed on any rejection and a status toast fired after the fact.

**Now rejected with an error**
- Renaming a text-based asset or folder to a name a sibling already uses (was previously only blocked for `.mjs`).
- Creating a `.js` or `.ts` script with a name already used in the folder (was previously only blocked for `.mjs`).
- Duplicating a script opens the create dialog seeded with the source filename so the user picks a unique name.

**Existing projects**
Pre-existing duplicates aren't migrated; they keep their names and load fine. Creating, renaming, or duplicating around them follows the new rules.

### Checks

### DIFF
--- a/sass/code-editor.scss
+++ b/sass/code-editor.scss
@@ -405,10 +405,15 @@ strong {
         }
 
         > .validate {
-            float: right;
-            margin: 2px 0 0;
+            margin: 0 0 8px;
             font-size: 12px;
             color: #f30;
+            white-space: normal;
+            word-break: break-word;
+
+            &:not(.pcui-hidden) {
+                display: block;
+            }
         }
     }
 }
--- a/sass/editor/_editor-main.scss
+++ b/sass/editor/_editor-main.scss
@@ -913,10 +913,17 @@ strong {
         }
 
         > .validate {
-            float: right;
-            margin: 2px 0 0;
+            margin: 0 0 8px;
             font-size: 12px;
             color: #f30;
+            white-space: normal;
+            word-break: break-word;
+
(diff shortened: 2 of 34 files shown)


Maksims:
The UX of this feature is not enough. Now renaming files can appear as broken as there is no proper UX related to it.
Also, many users won't use native VSCode extension, can they opt out of this forced behavior, and keep existing behavior?

lucaheft:
This is buggy, when copying from one project into another.
We have utility scripts which reside in a MadTools folder.
When copying the folder into a new project, the folder will be called MadTools (1).
This also happens, when copying single scripts.

kpal81xd:
> This is buggy, when copying from one project into another. We have utility scripts which reside in a MadTools folder. When copying the folder into a new project, the folder will be called MadTools (1). This also happens, when copying single scripts.

Fixed in #2048

PR #2061, opened 2026-05-22
fix: show material opacity intensity slider when dithering is enabled

lucaheft:
## Summary
Show the material opacity slider when opacity dithering or shadow dithering is enabled, in addition to the existing blend-type-based visibility.

## Problem
The opacity intensity slider in the material inspector was only shown when `blendType` was one of `[2, 4, 6]`. However, opacity intensity also has a visible effect when `opacityDither` or `opacityShadowDither` is set to anything other than `none` — meaning users could enable dithering but had no way to adjust the opacity value driving it.

## What's Changed
- Hook `toggleFields` into change events for `data.opacityDither` and `data.opacityShadowDither` so visibility updates when the user toggles dithering.
- Update the opacity field visibility check: the slider is now shown if the blend type matches **or** if either dither mode is not `none`.

## Test plan
- [x] With a non-blended blend type and both dither modes set to `none`, confirm the opacity slider is hidden.
- [x] Set `Opacity Dither` to a value other than `none` → slider becomes visible.
- [x] Reset to `none`, then set `Opacity Shadow Dither` to a non-`none` value → slider becomes visible.
- [x] Switch back to a blended blend type → slider remains visible regardless of dither settings.
- [x] Toggle dither values back and forth and confirm the slider shows/hides immediately without needing to re-open the inspector.

### Checks

### DIFF
--- a/src/editor/inspector/assets/material.ts
+++ b/src/editor/inspector/assets/material.ts
@@ -1424,6 +1424,8 @@ class MaterialAssetInspector extends Container {
         this._opacityInspector.getField('data.blendType').on('change', toggleFields);
         this._opacityInspector.getField('data.opacityVertexColor').on('change', toggleFields);
         this._opacityInspector.getField('data.opacityFadesSpecular').on('change', toggleFields);
+        this._opacityInspector.getField('data.opacityDither').on('change', toggleFields);
+        this._opacityInspector.getField('data.opacityShadowDither').on('change', toggleFields);
 
         this._clearCoatFactorInspector.getField('data.clearCoat').on('change', toggleFields);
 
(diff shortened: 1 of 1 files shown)


kpal81xd (changes requested):
Please address Copilot's inline comment before merge.

The current `dithered = opacityDither !== 'none' || opacityShadowDither !== 'none'` treats missing or empty legacy select values as dithered, so older non-blended materials can show the opacity intensity slider even when dithering is effectively off.

A future-proof fix should treat only truthy non-`none` values as enabled, so new dither modes still work while `undefined`, `null`, and `''` remain off.

I don't see any other PR-scoped blockers.

mvaligursky (approved):
Nice, thanks!

PR #2069, opened 2026-06-01
feat: expose gsplat rendering/LOD properties in component inspector

kpal81xd:
### What's Changed
- Expose `castShadows`, `lodBaseDistance` (min `0.1`, default `5`), and `lodMultiplier` (min `1.2`, default `3`) in the gsplat component inspector.
- Add reference tooltips with API reference links for the new fields, and backfill API reference links on the existing `asset`, `layers`, and component entries.

Requires the backend scene schema to include these gsplat properties so they receive defaults and persist; until that lands the fields drive the live engine entity in-session but won't serialize.

### Preview
image

### Checks

### DIFF
--- a/src/editor/attributes/reference/components/gsplat.ts
+++ b/src/editor/attributes/reference/components/gsplat.ts
@@ -4,15 +4,36 @@ export const fields: AttributeReference[]  = [{
     name: 'gsplat:component',
     title: 'pc.GSplatComponent',
     subTitle: '{pc.Component}',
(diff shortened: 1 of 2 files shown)


mvaligursky:
in the latest engine we've deprecated unified property, so we should not expose it. If the user really needs non-unified (which we will remove in the future version), they'd need to use script for this.

See the public API: https://api.playcanvas.com/engine/classes/GSplatComponent.html
we should not expose anything not public, for example High Quality SH

mvaligursky:
This is what we should expose:
- asset
- castShadows
- lodBaseDistance, lodMultiplier

PR #2078, opened 2026-06-08
feat: expose alphaDither in material inspector

lucaheft:
## Summary

Surfaces the engine's `StandardMaterial#alphaDither` property in the editor as a **Dither Intensity** slider under the material's Opacity section. This lets a material control the strength of opacity dithering independently of `opacity`, so it can be alpha-blended and dithered at the same time.

## Changes

- **Inspector** ([src/editor/inspector/assets/material.ts](src/editor/inspector/assets/material.ts)) — Adds a `Dither Intensity` slider bound to `data.alphaDither` (range `0`–`1`, step `0.05`). The slider is shown only when a dither mode is active (`opacityDither` or `opacityShadowDither` ≠ `none`) and the running engine supports `alphaDither`. Opacity-field visibility logic is updated so opacity and dither intensity coexist correctly.
- **Migration** ([src/editor/assets/assets-migrate.ts](src/editor/assets/assets-migrate.ts)) — Seeds `data.alphaDither` from `opacity ?? 1.0` on load when absent.
- **Reference docs** ([src/editor/attributes/reference/assets/material.ts](src/editor/attributes/reference/assets/material.ts)) — Adds the `asset:material:alphaDither` tooltip/reference entry linking to the engine API docs.
- **Types** ([src/editor-api/external-types/asset.d.ts](src/editor-api/external-types/asset.d.ts)) — Adds `alphaDither: number` to `MaterialAssetData`.

## ⚠️ Backend dependency

I've tested the changes with a locally built editor. However it seems like there are some backend changes needed as well.
The following warning is thrown when trying to change the dither intensity:

```
Error: invalid:exception: Invalid path
```

and the slider reverts to `0`.

Claude had the following remarks on this:
The realtime backend's material asset schema must whitelist the `data.alphaDither` path.
- **Path:** `data.alphaDither`
- **Type:** `number`
- **Range:** `0`–`1`


kpal81xd:
Ill add this field to the schema for you to verify

lucaheft:
Seems like the schema isn't updated yet. Any news on this?
We would like to make use of this in the editor.

kpal81xd:
> Seems like the schema isn't updated yet. Any news on this? We would like to make use of this in the editor.

Apologies for the wait - this schema is live now for you 🙂

lucaheft:
We might have a problem here.

What implications did adding this to the schema have? It seems like now every material defaults to alphaDither 1. Which would not preserve legacy behaviour. The default should be `null`, alphaDither getter properly handles fallback to opacity.

To reproduce:
Create a material with alphaFade set to 0.5.

Expected:
In the **current** editor this should create a material with half transparent half dithered material.
Result:
The material is half transparent with no dithering. (I assume alphaDither is set to 1 somewhere in the backend?) Confirmed with `editor.assets.get(295031903).get("data.alphaDither")` which returns `1`

This also introduces a problem: the alphaDither migration added in this PR never runs, because alphaDither is already defined.

kpal81xd:
Verified the bug - I’ve patched the backend to ensure default for alphaDither is set to null as intended.

lucaheft:
lgtm

I've adjusted the alphaDither migration to only set the alphaDither value to opacity if the material is dithered. Else 1.0 is used as default.

**Edit:**
Actually I'm not sure if setting alphaDither to null as default is the correct way. I guess the value could stay unset in the backend. (If this is supported?)
This would prevent the additional `asset.get('data.alphaDither') === null` check in the migration. Currently `asset.has('data.alphaDither')` will return true, because it's set to `null`

kpal81xd:
> lgtm
> 
> I've adjusted the alphaDither migration to only set the alphaDither value to opacity if the material is dithered. Else 1.0 is used as default.
> 
> **Edit:** Actually I'm not sure if setting alphaDither to null as default is the correct way. I guess the value could stay unset in the backend. (If this is supported?) This would prevent the additional `asset.get('data.alphaDither') === null` check in the migration. Currently `asset.has('data.alphaDither')` will return true, because it's set to `null`

Its better to have it set explicitly to null that omit it for schema consistency (implied defaults make migrations more complicated)

kpal81xd (approved):
LGTM rebase branch

PR #2130, opened 2026-06-29
refactor: automatically migrate engine v1 projects

kpal81xd:
Forum: https://forum.playcanvas.com/t/important-playcanvas-editor-support-for-engine-v1-ends-on-july-1st/42242

Engine V1 projects now move to Engine V2 automatically when a user with write access opens them.

### What's Changed

- Copies scene gamma correction and tone mapping to camera components, sets engineV2 to true, and reloads the Editor.
- Removes the manual Engine V1/V2 switch and its confirmation picker.
- Leaves WebGL 1, AudioSource, schema fields, inspectors, and other compatibility paths unchanged.

### Rollout

Deploy this frontend first. Writable Engine V1 projects migrate when opened. Read-only sessions do not persist a migration; a user with write access or the later backend migration must update those projects.

### Manual smoke test

1. Open a writable Engine V1 project with non-default scene gamma correction or tone mapping; expected: the camera values are updated, engineV2 becomes true, and the Editor reloads.
2. Open Settings > Engine; expected: the engine version selector remains, but the Engine V1/V2 switch is absent.
3. Open an existing Engine V2 project; expected: it opens without a migration or forced reload.
4. Open an Engine V1 project with read-only access; expected: no project or scene data is persisted.

### Checks


mvaligursky:
Just double checking. If user opens a project that has not been opened in years, we still handle V1 -> V2 migration for data, for example material properties?

Maksims:
What will happen to our V1 projects on our accounts?

mvaligursky:
> What will happen to our V1 projects on our accounts?

See here: https://forum.playcanvas.com/t/important-playcanvas-editor-support-for-engine-v1-ends-on-july-1st/42242/6

PR #1676, opened 2026-01-09
feat: add external code editor support with VS Code and Cursor deep linking

kpal81xd:
> [!IMPORTANT]
> ### Requirements
> - Asset collisions  https://github.com/playcanvas/vscode-extension/issues/77
> - Deep link to line - https://github.com/playcanvas/vscode-extension/pull/87
> - Backend schema - https://github.com/playcanvas/vscode-extension/pull/89

### What's Changed
- Added a new Editor setting - Code Editor
- When clicking edit on a text based asset for VSCode and Cursor will now deep link directly to opening the script in the extension
- Added a dismissible banner in the code editor prompting users to install the VSCode extension (localStorage-persisted dismissal)

### Testing
- [x] Open code editor from toolbar (no asset selected)
- [x] Open asset asset panel
- [x] Open asset from inspector panel
- [x] Error stack track link
- [x] VSCode extension banner appears at top of code editor
- [x] Dismiss button hides banner and persists across refreshes
- [x] Install Extension button opens marketplace in new tab
- [x] Banner hidden in file-only mode

### Preview
image

image

image

image

- [x] I confirm I have read the [contributing guidelines](https://github.com/playcanvas/editor/blob/main/.github/CONTRIBUTING.md)

PR #2189, opened 2026-07-30
feat: add Antigravity IDE external editor support

CuteTechGuy:
### What's Changed

  - Added **Antigravity IDE** (`antigravity-ide:`) as an external code editor option alongside VS Code and Cursor.
  - Added `Antigravity IDE` option to the Code Editor setting dropdown in user settings (`editor.codeEditor`).
  - Updated `assets:idePath` handlers to construct `antigravity-ide://` protocol URLs for script assets.
  - Added "Edit/View in Antigravity IDE" options to asset context menus and inspector dropdowns.
  - Updated launch error console and toolbar editor dispatchers to handle `antigravity-ide`.

  ### Checks

  ### Screenshots:
    
image

### DIFF
--- a/src/code-editor/assets/assets.ts
+++ b/src/code-editor/assets/assets.ts
@@ -198,7 +198,7 @@ editor.once('load', () => {
     editor.method('assets:getByVirtualPath', (path: string) => assetToVirtualPath.get(path));
 
     // get asset ide path
-    editor.method('assets:idePath', (ide: 'cursor' | 'vscode', asset?: Observer) => {
+    editor.method('assets:idePath', (ide: 'cursor' | 'vscode' | 'antigravity-ide', asset?: Observer) => {
         const assetPath = asset ? `/asset/${asset.get('id')}` : '';
         return `${ide}://playcanvas.playcanvas/project/${config.project.id}${assetPath}`;
     });
--- a/src/editor/assets/assets-context-menu.ts
+++ b/src/editor/assets/assets-context-menu.ts
@@ -583,6 +583,15 @@ editor.once('load', () => {
     });
     menu.append(menuItemEditCursor);
 
+    const menuItemEditAntigravity = new MenuItem({
+        text: `${writeLabel} in Antigravity IDE`,
+        icon: ICONS.EDIT,
+        onSelect: () => {
+            editor.call('assets:edit', currentAsset, 'antigravity-ide');
+        }
+    });
+    menu.append(menuItemEditAntigravity);
+
     // duplicate
     const menuItemDuplicate = new MenuItem({
         text: 'Duplicate',
@@ -738,6 +747,7 @@ editor.once('load', () => {
                 menuItemEditWeb.hidden = true;
                 menuItemEditVSCode.hidden = true;
                 menuItemEditCursor.hidden = true;
+                menuItemEditAntigravity.hidden = true;
                 if (editor.call('selector:type') === 'asset') {
                     const items = editor.call('selector:items');
                     menuItemDuplicate.hidden = items.length > 1 && items.indexOf(currentAsset) !== -1;
@@ -759,15 +769,18 @@ editor.once('load', () => {
                     menuItemEditWeb.hidden = editHidden;
                     menuItemEditVSCode.hidden = editHidden;
                     menuItemEditCursor.hidden = editHidden;
+                    menuItemEditAntigravity.hidden = editHidden;
                 } else {
                     menuItemEditWeb.hidden = false;
                     menuItemEditVSCode.hidden = false;
                     menuItemEditCursor.hidden = false;
+                    menuItemEditAntigravity.hidden = false;
                 }
             } else {
                 menuItemEditWeb.hidden = true;
                 menuItemEditVSCode.hidden = true;
                 menuItemEditCursor.hidden = true;
+                menuItemEditAntigravity.hidden = true;
             }
 
             // create atlas
@@ -951,6 +964,7 @@ editor.once('load', () => {
             menuItemEditWeb.hidden = true;
             menuItemEditVSCode.hidden = true;
             menuItemEditCursor.hidden = true;
+            menuItemEditAntigravity.hidden = true;
             menuItemDelete.hidden = true;
             menuItemReferences.hidden = true;
             menuItemReplace.hidden = true;
--- a/src/editor/assets/assets.ts
+++ b/src/editor/assets/assets.ts
@@ -124,7 +124,7 @@ editor.once('load', () => {
     });
 
     // get asset ide path
-    editor.method('assets:idePath', (ide: 'cursor' | 'vscode', asset?: AssetObserver) => {
+    editor.method('assets:idePath', (ide: 'cursor' | 'vscode' | 'antigravity-ide', asset?: AssetObserver) => {
         const assetPath = asset ? `/asset/${asset.get('id')}` : '';
         return `${ide}://playcanvas.playcanvas/project/${config.project.id}${assetPath}`;
     });
--- a/src/editor/inspector/asset.ts
+++ b/src/editor/inspector/asset.ts
@@ -417,6 +417,11 @@ class AssetInspector extends Container {
                     text: `${writeLabel} in Cursor`,
                     icon: 'E130',
                     onSelect: () => editor.call('assets:edit', this._assets[0], 'cursor')
+                },
+                {
+                    text: `${writeLabel} in Antigravity IDE`,
+                    icon: 'E130',
+                    onSelect: () => editor.call('assets:edit', this._assets[0], 'antigravity-ide')
                 }
             ]
         });
--- a/src/editor/inspector/settings-panels/editor.ts
+++ b/src/editor/inspector/settings-panels/editor.ts
@@ -277,6 +277,10 @@ const ATTRIBUTES: (Attribute | Divider)[] = [
                 {
                     v: 'cursor',
                     t: 'Cursor'
+                },
+                {
+                    v: 'antigravity-ide',
+                    t: 'Antigravity IDE'
                 }
             ]
         }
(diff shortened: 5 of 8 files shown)


src/editor/inspector/settings-panels/editor.ts
```diff
@@ -277,6 +277,10 @@ const ATTRIBUTES: (Attribute | Divider)[] = [
                 {
                     v: 'cursor',
                     t: 'Cursor'
+                },
+                {
+                    v: 'antigravity-ide',
+                    t: 'Antigravity IDE'
```
kpal81xd, line 283: This is the canonical definition site, and the value doubles as the URL protocol scheme, so it has to be `antigravity` to generate a link Antigravity actually handles.

```suggestion
                    v: 'antigravity',
                    t: 'Antigravity'
```

src/editor/assets/assets.ts
```diff
@@ -124,7 +124,7 @@ editor.once('load', () => {
     });
 
     // get asset ide path
-    editor.method('assets:idePath', (ide: 'cursor' | 'vscode', asset?: AssetObserver) => {
+    editor.method('assets:idePath', (ide: 'cursor' | 'vscode' | 'antigravity-ide', asset?: AssetObserver) => {
```
kpal81xd, line 127: Same rename, so the interpolated scheme is `antigravity://`.

```suggestion
    editor.method('assets:idePath', (ide: 'cursor' | 'vscode' | 'antigravity', asset?: AssetObserver) => {
```

src/launch/assets/assets.ts
```diff
@@ -86,7 +86,7 @@ editor.once('load', () => {
     });
 
     // get asset ide path
-    editor.method('assets:idePath', (ide: 'cursor' | 'vscode', asset) => {
+    editor.method('assets:idePath', (ide: 'cursor' | 'vscode' | 'antigravity-ide', asset) => {
```
kpal81xd, line 89: Same rename here.

```suggestion
    editor.method('assets:idePath', (ide: 'cursor' | 'vscode' | 'antigravity', asset) => {
```

src/code-editor/assets/assets.ts
```diff
@@ -198,7 +198,7 @@ editor.once('load', () => {
     editor.method('assets:getByVirtualPath', (path: string) => assetToVirtualPath.get(path));
 
     // get asset ide path
-    editor.method('assets:idePath', (ide: 'cursor' | 'vscode', asset?: Observer) => {
+    editor.method('assets:idePath', (ide: 'cursor' | 'vscode' | 'antigravity-ide', asset?: Observer) => {
```
kpal81xd, line 201: Same rename here. Side note for point 3 in my summary: this union is now hand-copied across three files, which is exactly the kind of thing that drifts. A single exported type would keep them honest.

```suggestion
    editor.method('assets:idePath', (ide: 'cursor' | 'vscode' | 'antigravity', asset?: Observer) => {
```

src/editor/toolbar/toolbar-code-editor.ts
```diff
@@ -25,7 +25,7 @@ editor.once('load', () => {
         (asset?: Observer, options?: Record<string, unknown>, popup?: boolean, ideOverride?: string) => {
             // open the code editor external editor
             const ide = ideOverride || projectUserSettings.get('editor.codeEditor');
-            if (ide === 'vscode' || ide === 'cursor') {
+            if (ide === 'vscode' || ide === 'cursor' || ide === 'antigravity-ide') {
```
kpal81xd, line 28: Same rename. This guard and the one in `viewport-error-console.ts` are now both three-way ORs that have to stay in sync; a shared `EXTERNAL_IDES` constant with `.includes(ide)` would collapse both and make the next IDE a one-line change.

```suggestion
            if (ide === 'vscode' || ide === 'cursor' || ide === 'antigravity') {
```

src/launch/viewport/viewport-error-console.ts
```diff
@@ -115,7 +115,8 @@ editor.once('load', () => {
                     assetId = parseInt(match[1], 10);
                     switch (ide) {
                         case 'vscode':
-                        case 'cursor': {
+                        case 'cursor':
+                        case 'antigravity-ide': {
```
kpal81xd, line 119: Same rename.

```suggestion
                        case 'antigravity': {
```

src/launch/viewport/viewport-error-console.ts
```diff
@@ -157,7 +158,7 @@ editor.once('load', () => {
             if (assetId) {
                 const link = document.getElementById(`error-${errorCount}`);
                 link.addEventListener('click', (e: MouseEvent) => {
-                    if (ide === 'vscode' || ide === 'cursor') {
+                    if (ide === 'vscode' || ide === 'cursor' || ide === 'antigravity-ide') {
```
kpal81xd, line 161: Same rename. Note this guard has to match the `switch` above it, since a mismatch would build the external-IDE URL and then fall through to the tab-reuse path.

```suggestion
                    if (ide === 'vscode' || ide === 'cursor' || ide === 'antigravity') {
```

src/editor/assets/assets-context-menu.ts
```diff
@@ -583,6 +583,15 @@ editor.once('load', () => {
     });
     menu.append(menuItemEditCursor);
 
+    const menuItemEditAntigravity = new MenuItem({
+        text: `${writeLabel} in Antigravity IDE`,
+        icon: ICONS.EDIT,
+        onSelect: () => {
+            editor.call('assets:edit', currentAsset, 'antigravity-ide');
```
kpal81xd, line 590: Rename plus the shorter label.

```suggestion
        text: `${writeLabel} in Antigravity`,
        icon: ICONS.EDIT,
        onSelect: () => {
            editor.call('assets:edit', currentAsset, 'antigravity');
```

src/editor/inspector/asset.ts
```diff
@@ -417,6 +417,11 @@ class AssetInspector extends Container {
                     text: `${writeLabel} in Cursor`,
                     icon: 'E130',
                     onSelect: () => editor.call('assets:edit', this._assets[0], 'cursor')
+                },
+                {
+                    text: `${writeLabel} in Antigravity IDE`,
+                    icon: 'E130',
+                    onSelect: () => editor.call('assets:edit', this._assets[0], 'antigravity-ide')
```
kpal81xd, line 424: Rename plus the shorter label here too.

```suggestion
                    text: `${writeLabel} in Antigravity`,
                    icon: 'E130',
                    onSelect: () => editor.call('assets:edit', this._assets[0], 'antigravity')
```

kpal81xd (changes requested):
Thanks for this, and nice job on the coverage. You found every site that branches on the IDE value: all three `assets:idePath` registrations, the launch error-console switch *and* its click handler, the toolbar guard, the settings dropdown, and all four context-menu visibility toggles. I went looking for a missed dispatch site and there isn't one.

Two things need to change before it can work, though, plus one thing I'd like to settle first.

### 1. The URL scheme is wrong, so the deep link silently does nothing

The setting value is interpolated directly as the protocol scheme, so `antigravity-ide` produces `antigravity-ide://playcanvas.playcanvas/...`:

https://github.com/playcanvas/editor/blob/78d7d576c3d3f7002aef07a98f0e4fb6e3ffb1a3/src/editor/assets/assets.ts#L126-L129

Antigravity registers `antigravity://`, not `antigravity-ide://` (its own extension deep links use the bare `antigravity:extension/<id>` form). An unregistered scheme fails silently in the browser: no error, no IDE, the click just does nothing. Renaming the value to `antigravity` fixes it, and I've left commit-ready suggestions on every occurrence. I've also suggested dropping "IDE" from the user-visible labels, since the product is just called Antigravity.

Could you verify against a real install before we merge? A screenshot of a script opening at the right line in Antigravity would be ideal. Worth confirming the extension resolves there too: it is published on Open VSX (the registry Antigravity ships with) so it should, but the URL hardcodes the `playcanvas.playcanvas` identifier and the Open VSX namespace is capitalised, so please sanity-check that the handler actually fires rather than just the IDE coming to the foreground.

### 2. This needs the paired backend schema change

`editor.codeEditor` is validated server-side against a fixed set of allowed values, so choosing the new option in the Editor settings panel will be rejected and won't persist. This is the same requirement #1676 carried in its checklist. The context-menu and inspector entries would still work, since those pass the IDE explicitly rather than reading the setting, but the dropdown wouldn't. That change isn't something you can make from this repo, so we'll handle it on our side.

### 3. Worth settling: this doesn't scale to the next fork

Each new VS Code fork now costs an eight-file change here plus a backend update, and Windsurf, Trae and Kiro are all plausible next requests. Two shapes I'd prefer: hoist the IDE list into a single shared constant with a derived type, so the next addition touches one file instead of eight; or drop the enum entirely in favour of one "custom URL scheme" text field that covers every fork permanently.

To be clear, that isn't a criticism of this PR, which faithfully follows the pattern #1676 established. But I'd rather resolve it before we add a third hardcoded entry. Happy to take either direction here, or to land the rename first and refactor separately. Let me know which you'd prefer.

CuteTechGuy:
Just a quick question, do you use Mac Os or Windows ? With Antigravity 2.0 google split the ide and agent part. Right now there is two different app in Mac OS, Antigravity 2.0(Agent Interface) and Antgiravity IDE. Thats why I used the `antigravity-ide:` prefix  because `antigravity:` forwards to the agent app and does nothing 😢

kpal81xd:
> Just a quick question, do you use Mac Os or Windows ? With Antigravity 2.0 google split the ide and agent part. Right now there is two different app in Mac OS, Antigravity 2.0(Agent Interface) and Antgiravity IDE. Thats why I used the `antigravity-ide:` prefix because `antigravity:` forwards to the agent app and does nothing 😢

Ahh thats bad design but yes I see the split - I was wondering where there ide bit was when downloading for Mac :)

kpal81xd:
Once you are ready just request me to review the PR for you @CuteTechGuy

kpal81xd:
@CuteTechGuy whats the status of this? I can take over this PR if you wish to get it merged

CuteTechGuy:
@kpal81xd Hi! Thanks for checking in and offering to help. I'm really sorry for the late reply; I've been away for a while dealing with some personal matters.

Please feel free to take over the PR and finish it up to get it merged. I appreciate the help!

Issue #2225, opened 2026-08-31
Sticky Tooltips, prevent from UI interactions


Maksims:
Often some tooltips stick around, and they intercept pointer events.
And when they stick over some important buttons, you wont be able to interact with buttons.

Solution:
1. Make sure tooltips dissapear.
2. Small tooltips, should have CSS `pointer-events: none` on them, to ensure they do not intercept pointer events.

<img width="455" height="321" alt="Image" src="https://github.com/user-attachments/assets/d0ad0df1-576a-4957-9e1b-7dc2aaea3380" />

PR #2229, opened 2026-09-02
fix: prevent simple tooltips from intercepting pointer events

shin1244:
Partially addresses #2225 (item 2)

### What's Changed

- Added `pointer-events: none` to `.tooltip-simple` in SCSS.
- Prevents simple tooltips from intercepting pointer events and blocking UI interactions underneath.
- Scoped strictly to `.tooltip-simple` so interactive tooltips (`.tooltip-reference`, `.tooltip-override`) containing buttons remain clickable.

### Checks

### DIFF
--- a/sass/editor/_editor-tooltip.scss
+++ b/sass/editor/_editor-tooltip.scss
@@ -9,6 +9,7 @@
     position: static;
     max-width: 300px;
     flex-shrink: 0;
+    pointer-events: none;
 
     > .pcui-label {
         white-space: normal;


Maksims:
There are cases where tooltip should be interactible, for example: API references, they can have buttons/links.
{008D4FFE-5CE2-4679-B003-B78DB41E3908}

shin1244:
Verified locally, the API REFERENCE button remains clickable.
It looks like .tooltip-reference is separate from .tooltip-simple, so it remains unaffected.

kpal81xd (approved):

PR #2237, opened 2026-09-03
Feat:assets split view

### LINKED ISSUE #2232: Split view for the assets panel
## Split View for the Assets Panel

Our team added a split view to the assets panel. In a project with a few thousand assets, moving or comparing files across folders meant navigating back and forth constantly.

A **SPLIT** button opens a second panel beside the first. Each panel has its own folder tree, view mode, and current folder, so assets can be compared or dragged directly from one folder into another.

Split View

The active asset panel is highlighted with an orange border.

### PR

I have a working implementation of this feature based on the current editor codebase.

Would this feature be considered suitable for the project? If so, I can open a pull request with the implementation.

shin1244:
Fixes #2232

### What's Changed

- WIP: Split view for the Assets panel.
- Still working on the implementation and will address the remaining changes discussed in the issue.
- Please let me know if there are any other changes you'd like me to make.

### Checks

### DIFF
--- a/sass/editor/_editor-assets-panel.scss
+++ b/sass/editor/_editor-assets-panel.scss
@@ -613,21 +613,45 @@
 
 // split view - two asset panels side by side
 #layout-assets-dock.assets-split {
-    #layout-assets {
-        border-right: 1px solid $bcg-darkest;
+    // pcui indents every panel header by 10px, which would leave a gap between the divider
+    // and the marker below. Just clear the divider instead
+    #layout-assets-secondary > .pcui-panel-header {
+        padding-left: 4px;
     }
 
-    // make it obvious which of the two panels new assets and hotkeys apply to
-    .pcui-asset-panel-active > .pcui-panel-content::after {
-        content: '';
-        position: absolute;
-        top: 0;
-        right: 0;
-        left: 0;
-        height: 2px;
+    // the secondary view does not repeat the panel title. The slot it leaves behind is
+    // what marks the view that menu actions apply to - the main view needs no marker
+    #layout-assets-secondary > .pcui-panel-header > .pcui-panel-header-title {
+        width: 6px;
+        padding: 0;
+        margin: 0;
+        font-size: 0;
+    }
+
+    #layout-assets-secondary.pcui-asset-panel-active > .pcui-panel-header > .pcui-panel-header-title {
         background-color: $text-active;
-        pointer-events: none;
-        z-index: 1;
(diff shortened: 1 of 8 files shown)


shin1244:
@Maksims 
Done. The left view is the main one now and navigation always goes there, so the
right view keeps whatever the user put in it. Pasting is resolved from the pointer,
and the "active view" idea is down to a small marker in the right view's header.

There is one case I would like your view on. The setup in the video is deliberately
extreme, but copying assets between projects with several editor windows open is a
real workflow for us.

[녹음 2026-09-04 112008.webm](https://github.com/user-attachments/assets/67ed3e64-ab21-43de-98e8-e0cbd47e7dcc)

1. The user activates the Hovership folder view, intending pasted assets to land there.
2. The window they copy from overlaps this one, so coming back means clicking on
   whatever part is visible — here, the hierarchy.
3. The pointer is now over the hierarchy, so the paste lands in the main view, and the
   assets have to be moved into Hovership afterwards.

I think an explicit "active view" concept could help avoid this kind of ambiguity. That said, I'm not a UX expert, so I'd appreciate your perspective on whether this would be a good UX approach.

shin1244:
> I think an explicit "active view" concept could help avoid this kind of ambiguity. That said, I'm not a UX expert, so I'd appreciate your perspective on whether this would be a good UX approach.

To add to that — what I mean by an active view is one simple rule: everything happens
in the panel you clicked last.

I expect that means less code and a more intuitive result, with no special cases to
write for each action.

Maksims:
It is a complex UX indeed.

In this case, it would be better if it would paste in the right panel, indeed. So for indirect pasting - active panel - is a needed concept.

But here is a case where applying "active panel" to "auto-navigation", would be pretty annoying:
1. User navigates to folder with textures on a second view.
2. User selects render from viewport.
3. User selects material from an inspector.
4. We want auto-navigation to always happen on the left view, so that right view with textures is readily available.

If we auto-navigate based on active panel, then case with multiple materials selected and texture folder navigations, will lead to ping-pong auto-navigations, and will require an extra deliberate focus of the left or right view by user to ensure "sticky" auto-navigation. Which is undesirable.

So for auto-navigation - we want sticky one (on the main/left view), while for other actions (without direct interaction) - based on active, highlighted view.

shin1244:
Understood, changed it that way.

Auto-navigation is pinned back to the main view, while pasting and asset creation
follow the active view.

PR #2238, opened 2026-09-04
feat: align Editor writes with complete ShareDB schemas

kpal81xd:
### What's Changed

- Serialize entity/template linkage fields explicitly, using `null` when unset instead of omitting or deleting the keys.
- Sync the `template` field in Editor and Launch, and apply incoming camera user-data updates.
- Keep document metadata out of settings updates without mutating the ShareDB snapshot.
- Add regression coverage for entity/template serialization and backend updates during pending edits, including undo/redo.

Backfill belongs to the accompanying backend changes: when an existing document is edited, the backend fills missing schema-defined defaults without overwriting existing values. Invalid values still fail validation. This PR contains no frontend repair/retry coordinator, bulk migration, or read-triggered backfill.

### Checks

- [x] CI build, lint, and tests pass at `a9d9e9174`
- [x] Test Suite CI typecheck and lint pass at `a9d9e9174`

### Manual smoke test

Run with the corresponding backend schema enforcement/backfill changes.

1. Create an entity, copy/paste it, and create a template from it. Reload and confirm the hierarchy and template links are preserved without validation errors.
2. Open an existing scene with missing defaulted keys in two Editor windows and rename an entity. Expect both windows to show the saved edit, with only missing defaults added by the backend.
3. Undo and redo the rename. Expect only the user edit to change; the backfilled defaults must remain and must not create a separate undo action.
4. Change project settings and move the editor camera, then reload. Expect the settings and camera state to persist without synchronization errors.


Maksims:
May I ask why repairing is required?

Breaking user data in first place should not be possible due to schema validations on the back-end.

kpal81xd:
> May I ask why repairing is required?
> 
> Breaking user data in first place should not be possible due to schema validations on the back-end.

Repair is the wrong word - its a backfill of missing schema defined defaults to follow the stricter validation i have in place. It does not change or overwrite any existing data. Also this should never have been in the frontend - its been moved to the backend

